PR #4, opened 2021-02-12
Guess Who, Ylva Landoff Lindberg

sssylvasss:


code/script.js
```diff
@@ -255,28 +249,25 @@ const generateBoard = () => {
 // Randomly select a person from the characters array and set as the value of the variable called secret
 const setSecret = () => {
   secret = charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)]
-  console.log(secret)
```
leelalaim: you can delete the console.log after it is no more needed

leelalaim: you can delete the console.log after it is no more needed.

code/script.js
```diff
@@ -291,54 +282,44 @@ const selectQuestion = () => {
...
-    currentQuestion = {
-      attribute: 'glasses',
-      value: true, 
-      category
-    }  
+      currentQuestion = {
+        attribute: 'glasses',
+        value: true, 
+        category
+      }
```
leelalaim: To make a bit less code, you can combine categories "glasses" and "hat" into single category called "accessories", and then : 
`currentQuestion = {
      attribute: value,
      // 👆 this is the property of the booleans such as smoke, glasses and hat. add the value from the input here
      value: true, // we're asking if this person wears a hat for example, so always true in the question.
      category: category,
    }`

leelalaim: To make a bit less code, you can combine categories "glasses" and "hat" into single category called "accessories", and then :
`
currentQuestion = {
attribute: value,
value: true,
category: category,
}
`
Attribute has a property of the booleans such as glasses or hat.
In the "value" we are asking if this person wears a hat for example, so it always has to be true in the question.

code/script.js
```diff
@@ -291,54 +282,44 @@ const selectQuestion = () => {
...
-
 const checkQuestion = () => {
-
   selectQuestion()
   const secretValue = secret[currentQuestion.attribute]
   if (secretValue === currentQuestion.value) {
     filterCharacters(true,currentQuestion.category);
   } else {
     filterCharacters(false,currentQuestion.category);
   }
```
leelalaim: Here, for less code, you can use use a Ternary Operator:
`secretValue === currentQuestion.value ? filterCharacters(true) : filterCharacters(false)`

code/script.js
```diff
@@ -399,39 +379,34 @@ const filterCharacters = (keep) => {
...
     winOrLoseText.innerHTML = `Yey! High Five! It was ${secret.name}`
     winOrLose.style.display = 'block';
   } else {
     winOrLose.style.display = 'block';
     winOrLoseText.innerHTML = `Sorry! It is not ${suspect}`
   }
-
     winOrLose.style.display = 'flex'
     board.style.display = 'none'
 };
```
leelalaim: Nice and neat code, well done! 💯

leelalaim (commented):
Nice code. Well done! :) I have added some comments

code/index.html
```diff
@@ -26,16 +26,25 @@ <h1>Does the person have</h1>
...
-          <!-- Add the other accessory option here. (hat)  -->
+        <optgroup label="hat">
+          <option value="hat">hat</option>
         </optgroup>
+        <optgroup label="glasses">
+          <option value="glasses">glasses</option>
+        </optgroup>  
         <optgroup label="other">
           <option value="smoker">a smoking habit</option>
         </optgroup>
```
KarolCislowski, line 50: Maybe it would be a nice think to generate long lists of options with your Javascript? It is nothing wrong, with this approach, but It makes you type a lot 😸

code/scripts/characters.js
```diff
@@ -0,0 +1,220 @@
+// Array with all the characters, as objects
```
KarolCislowski, line 1: Nice that you tried to export your data to the external file 🌈 why didn't you use it in your script? 🤓

KarolCislowski (approved):
This is a very nice project 🌈 

Remember that you can always come back to it when you have some spare time and add more functionalities to make it even greater portfolio asset! 🦄

code/script.js
sssylvasss: When it's red it already delete :)

code/scripts/characters.js
sssylvasss: I can't make it work, don't connect:(

PR #10, opened 2021-02-12
Guess Who

Aepera:


code/script.js
```diff
@@ -225,9 +228,10 @@ const CHARACTERS = [
 ]
 
 // Global variables
-let secret, currentQuestion, charactersInPlay
+let secret, currentQuestion, charactersInPlay, secretValue
+let startNumber = 0;
```
CarolineF3, line 232: The name of the variable could be a little bit more descriptive. For example: numberOfGuesses

code/script.js
```diff
@@ -244,56 +248,72 @@ const generateBoard = () => {
...
 
-// This function should be invoked when you click on 'Find Out'.
 const checkQuestion = () => {
-  // Compare the currentQuestion with the secret person.
-  // See if we should keep or remove people based on that
-  // Then invoke filterCharacters
+  // Compares the currentQuestion with the secret person.
+  const secretValue = secret[currentQuestion.attribute];
+
+  if (currentQuestion.value === secretValue) {
```
CarolineF3, line 302: This variable isn't necessary, it's possible to write the if statement like this: `if (currentQuestion.value === secret[currentQuestion.attribute])`.  On the other hand, the variable provides clarity, so it might be good for the readability (and worth keeping). :)

code/script.js
```diff
@@ -244,56 +248,72 @@ const generateBoard = () => {
...
+    keep = true
+    filterCharacters(true)
+  } else {
+    keep = false
+    filterCharacters(false)
+  }
```
CarolineF3, line 308: Instead of making the function call for filterCharacter inside the if statement you could make the function call after the if statement and pass the keep variable as a function argument. Example: filterCharacters(keep).

code/script.js
```diff
@@ -303,41 +323,92 @@ const filterCharacters = (keep) => {
...
+    board.innerHTML = ''
+    document.getElementById('winOrLose').style.display = 'flex'
+    numberOfGuesses.innerHTML = `You guessed ${startNumber} times`
+    showSecretCard.innerHTML = `
+      <img src=${secret.img}>
+      <h2>${secret.name}</h2>
+      `
+  } else {
+    alert(`No, this is not the secret person, keep guessing!`)
+  }
```
CarolineF3, line 390: This is a nice take on the game! :)

code/script.js
```diff
@@ -303,41 +323,92 @@ const filterCharacters = (keep) => {
...
 }
 
 // Invokes the start function when website is loaded
 start()
 
-// All the event listeners
+const clickCounter = () => {
+  // This function counts how many times the player clicks on Find Out-button.
+  startNumber = 1+startNumber++
+}
```
CarolineF3, line 399: Nice job on putting the counter inside its own function! :)

CarolineF3:
Really great job, the code looks very clear and tidy! It's easy to comprehend. You meet the requirements and the guess counter is a nice extra feature. :)

CarolineF3 (approved):

code/script.js
KarolCislowski: It is a good practice to create a short-named variable to store longer chains of data before it will be passed to the logic. It makes it extremely easier in larger projects. sometimes one extra line of code can save you houers of debugging 😸

code/index.html
```diff
@@ -26,22 +26,32 @@ <h1>Does the person have</h1>
...
         </optgroup>
         <optgroup label="accessories">
           <option value="glasses">glasses</option>
-          <!-- Add the other accessory option here. (hat)  -->
+          <option value="hat">hat</option>
+
         </optgroup>
         <optgroup label="other">
           <option value="smoker">a smoking habit</option>
         </optgroup>
```
KarolCislowski, line 51: Maybe it would be a nice think to generate long lists of options with your Javascript? It is nothing wrong, with this approach, but It makes you type a lot 😸

code/script.js
```diff
@@ -303,41 +323,92 @@ const filterCharacters = (keep) => {
...
+  } else {
+  }
 }
 
-// If you confirm, this function is invoked
 const checkMyGuess = (suspect) => {
-  // 1. Check if the suspect is the same as the secret person's name
-  // 2. Set a Message to show in the win or lose section accordingly
-  // 3. Show the win or lose section
-  // 4. Hide the game board
```
KarolCislowski, line 336: Great that you remembered to remove comments with suggestions from the coaches 🥳

KarolCislowski (approved):
This is a very nice project 🌈 

PR #15, opened 2021-02-13
Project 5 - Isabella Billgren

Bellsibub:
!!See ReadMe for project info!!
Link to published project: https://guess-who-flips.netlify.app/


README.md
```diff
@@ -1,13 +1,125 @@
...
+  background-color: transparent;
+  color: var(--primary);
+}
+.button__filled {
+  background-color: var(--primary);
+  color: white;
+}
+```
+It is an elegant way of making sure that all buttons have a base style (applied by the **.button** class).
+Then the element has another class on it to make sure it has a unique style to it (in this case applied by the **.button__outlined** and **.button__filled**)
```
nazi-pira, line 63: What a smart idea!

code/scripts/cards.js
```diff
@@ -0,0 +1,21 @@
...
+                    <button class="button button__filled button__filled--small" id="guess" data-name="${person.name}">Guess</button>
+                </div>
+            </div>
+            <div class="card__back">
+                <img class="icon icon--card-back" src="images/guess-who-bubble.png" alt="Guess Who" />
+            </div>
+        </div>    
+    </div>
+    `;
+  return html;
```
nazi-pira, line 20: Does the return put the card back to its default form after being clicked?

code/style.css
```diff
@@ -1,233 +1,15 @@
-/* Global css variables used for colors */
-:root {
-  --primary: #a259ff;
```
nazi-pira, line 3: What does these two -- do? :D

code/styles/buttons.css
```diff
@@ -0,0 +1,36 @@
...
+  border: 2px solid var(--primary);
+  border-radius: 50px;
+  padding: 17px 50px;
+  margin: 10px 0;
+  cursor: pointer;
+}
+
+/* class to make an outline button */
+.button__outlined {
+  background-color: transparent;
```
nazi-pira, line 14: Couldn't we make the button absolute instead of transparent background?

code/styles/cards.css
```diff
@@ -0,0 +1,84 @@
...
+  height: 100%;
+  transition: all 0.8s ease-in-out;
+  transform-style: preserve-3d;
+  border-radius: 4px;
+  border: 2px solid var(--secondary);
+}
+
+/* Flipped state of the inner wrapper
+  (the card flips when it has this class) */
+.card__inner--flipped {
```
nazi-pira, line 24: a very impressive feature!

nazi-pira (approved):
The code seemed to be so clean and professional to me!

README.md
```diff
@@ -1,13 +1,125 @@
...
+
+***
+
+## Workflow
+
+### The steps
+I began by reading through the steps defined in the project brief. I then used those descriptions in conjunction with the listed blue requirements to create a todo list of functions to add. I worked through the list and added functionality accordingly. 
+
+### Make it BEM
+Like I've done in almost all my projects so far I go through all the html elements and existing style sheets. I then change and apply the BEM methodology to that.
```
KarolCislowski, line 35: Great methodology! it makes your stylesheets so easy to maintain 😻

README.md
```diff
@@ -1,13 +1,125 @@
...
+If I were to continue on this project / start over I would:
+- I would add a custom confirm alert (since I already made a custom normal alert, I would probably add the same type of thing for the confirm action)
+- I would also probably work more on the style for the question selection.
+  - Maybe even make so that you can't guess the same thing twice.
+<br>
+
+***
 
 ## View it live
+https://guess-who-flips.netlify.app/
```
KarolCislowski, line 125: Your ReadMe files are screaming "READ ME AGAIN AND AGAIN" 😻  and this is strange because I do not like to read 🙈

code/index.html
```diff
@@ -3,62 +3,99 @@
...
+            <option value="necklace">a necklace</option>
+            <option value="earrings">earrings</option>
+            <option value="headCover">a head covering</option>
+          </optgroup>
+          <optgroup label="other">
+            <option value="smoker">a smoking habit</option>
+            <option value="facialHair">facial hair</option>
+            <option value="buttonShirt">a buttoned up shirt</option>
+          </optgroup>
+        </select>
```
KarolCislowski, line 56: Maybe it would be a nice think to generate long lists of options with your Javascript? It is nothing wrong, with this approach, but It makes you type a lot 😸

code/scripts/cards.js
KarolCislowski: No return is the keyword which is describing what function will send outside after all calculations happened.

code/scripts/cards.js
```diff
@@ -0,0 +1,21 @@
+// Function to dynamically return a card template based on character object data
+export const getCard = (person) => {
+  let html = `
```
KarolCislowski, line 3: You are creating it once, and then you do not change it so it should be created as a `const`

code/scripts/characters.js
```diff
@@ -0,0 +1,315 @@
+export const CHARACTERS = [
```
KarolCislowski, line 1: External Dataset... so clean 😻

code/scripts/characters.js
```diff
@@ -0,0 +1,315 @@
...
+    img: "images/jabala.svg",
+    hairColor: "hidden",
+    eyeColor: "hidden",
+    glasses: true,
+    necklace: false,
+    earrings: false,
+    headCover: true,
+    smoker: false,
+    facialHair: false,
+    buttonShirt: false,
```
KarolCislowski, line 13: extra fields 🦄

KarolCislowski (approved):
Another Masterpiece 🦄 

Remember to create your variables by using `const` and change it to `let` when the compiler is yelling at you that you can change the constant 🤓

PR #18, opened 2021-02-13
Project 5: Guess who?

Andrealarsson:


Pauan8 (changes requested):
Well done, really nice and structured code 🥇! I just have some tiny changes needed. 

**Changes Requested**
-Remove unused variables!

code/script.js
```diff
@@ -244,100 +247,165 @@ const generateBoard = () => {
...
+    keep = true;
+    filterCharacters(true, currentQuestion.attribute)
+  } else {
+    keep = false
+    filterCharacters(false, currentQuestion.attribute)
+  }
 }
 
 // It'll filter the characters array and redraw the game board.
 const filterCharacters = (keep) => {
```
KarolCislowski, line 315: You are creating your `keep` here and you are using it inside. there is no need to create a global `keep` because it is never read.

KarolCislowski (approved):
This is a very nice project 🌈 

PR #22, opened 2021-02-13
Week 5 - Guess Who - Sofia A.W

Sofarz:
Finished the project, did not do any of the stretch goals.

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -254,6 +254,8 @@ const start = () => {
   // Here we're setting charactersInPlay array to be all the characters to start with
   charactersInPlay = CHARACTERS
   // What else should happen when we start the game?
+  setSecret()
+  generateBoard()
 }
 
 // setting the currentQuestion object when you select something in the dropdown
--- a/code/script.js
+++ b/code/script.js
@@ -2,6 +2,9 @@
 const board = document.getElementById('board')
 const questions = document.getElementById('questions')
 const restartButton = document.getElementById('restart')
+const filterBtn = document.getElementById('filter');
+const winOrLose = document.getElementById('winOrLose');
+const playAgainBtn = document.getElementById('playAgain');
 
 // Array with all the characters, as objects
 const CHARACTERS = [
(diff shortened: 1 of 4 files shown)


code/script.js
```diff
@@ -2,6 +2,10 @@
 const board = document.getElementById('board')
```
M4L5, line 2: In this area some of the DOM's have semicolons. Just a heads up to double check and choose if you want to go with semicolons or not. I actually did the same in my code so I will also need to double check my own :)

code/script.js
```diff
@@ -247,97 +252,160 @@ const generateBoard = () => {
...
+  console.log('secret value: ', secretValue);
+  if (secretValue === currentQuestion.value) {
+    filterCharacters(true);
+  } else 
+    filterCharacters(false);
 }
 
 // It'll filter the characters array and redraw the game board.
 const filterCharacters = (keep) => {
+  let category = currentQuestion.category
```
M4L5, line 314: This part is very clear and and the code is easy to read. :D

M4L5 (commented):
An overall comment for the whole code is that I think it looks good. It's easy to read and follow. I can't see anything right now that I can give suggestions on how to do in another way or think about. I think you did a great job with everything going on this week. Someone is a cool coding mom! 👍 👍

code/script.js
KarolCislowski: You need to check your VScode settings and search for semicolons and if you decide to have them then make Prettier(vscode's extension- if you don't have it you need to install it) add them automatically after each saves.

code/script.js
```diff
@@ -247,97 +252,160 @@ const generateBoard = () => {
...
+  let attribute = currentQuestion.attribute
+  let value = currentQuestion.value
+
+  console.log('category: ', category);
```
KarolCislowski, line 318: remember to always clean your code from all console.log()s and commented out code blocks 🤓

KarolCislowski (approved):
This is a very nice project 🌈 

PR #23, opened 2021-02-13
Guess Who Hannah

jesinkey:
Have fun!

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -4,224 +4,223 @@ const questions = document.getElementById('questions')
 const restartButton = document.getElementById('restart')
 
 // Array with all the characters, as objects
-const CHARACTERS = [
-  {
-    name: 'Jabala',
-    img: 'images/jabala.svg',
-    hairColor: 'hidden',
-    eyeColor: 'hidden',
-    glasses: true,
-    hat: true,
-    smoker: false,
-  },
-  {
-    name: 'Jack',
-    img: 'images/jack.svg',
-    hairColor: 'hidden',
-    eyeColor: 'blue',
-    glasses: false,
-    hat: true,
-    smoker: false,
-  },
-  {
-    name: 'Jacques',
-    img: 'images/jacques.svg',
-    hairColor: 'grey',
-    eyeColor: 'blue',
-    glasses: false,
-    hat: true,
-    smoker: true,
-  },
-  {
-    name: 'Jai',
-    img: 'images/jai.svg',
-    hairColor: 'black',
-    eyeColor: 'brown',
-    glasses: false,
-    hat: false,
-    smoker: false,
-  },
-  {
-    name: 'Jake',
-    img: 'images/jake.svg',
-    hairColor: 'yellow',
-    eyeColor: 'green',
-    glasses: true,
-    hat: false,
-    smoker: false,
-  },
-  {
-    name: 'James',
-    img: 'images/james.svg',
-    hairColor: 'brown',
-    eyeColor: 'green',
-    glasses: true,
-    hat: false,
-    smoker: false,
-  },
-  {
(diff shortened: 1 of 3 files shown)


code/script.js
```diff
@@ -2,237 +2,239 @@
...
+            name: 'Jabala',
+            img: 'images/jabala.svg',
+            hairColor: 'hidden',
+            eyeColor: 'hidden',
+            glasses: true,
+            hat: true,
+            smoker: false,
+        },
+        {
+            name: 'Jack',
```
jessika-hage, line 21: Maybe the indentation could be fixed here, but don't know if that is necessary in this case! It still looks clean!

code/script.js
```diff
@@ -241,103 +243,152 @@ const generateBoard = () => {
...
-      // value: ,
-      // 👆 add the value from the input here
-      category: category,
+    /* the variable currentQuestion is created as an object to use for comparison with the "secret"-person*/
+    if (category === 'hair color') {
+        currentQuestion = {
+            attribute: 'hairColor',
+            value: selectedValue,
+            category: category,
+        }
```
jessika-hage, line 277: When it's the same attribute as value, you could just write category, :)

code/script.js
```diff
@@ -241,103 +243,152 @@ const generateBoard = () => {
...
+    generateBoard(charactersInPlay)
 }
 
-// when clicking guess, the player first have to confirm that they want to make a guess.
+/*the player confirms that they are making a finalGuess*/
 const guess = (suspect) => {
-  // store the interaction from the player in a variable.
-  // remember the confirm() ?
-  // If the player wants to guess, invoke the checkMyGuess function.
+    let finalGuess = confirm(`Are you sure you would like guess ${suspect}?`)
```
jessika-hage, line 370: You could use const instead of let since you're not reassigning it!

code/script.js
```diff
@@ -241,103 +243,152 @@ const generateBoard = () => {
...
 }
 
+
 // setting the currentQuestion object when you select something in the dropdown
 const selectQuestion = () => {
-  const category = questions.options[questions.selectedIndex].parentNode.label
-  // This variable stores what option group (category) the question belongs to.
-  // We also need a variable that stores the actual value of the question we've selected.
+    const category = questions.options[questions.selectedIndex].parentNode.label
+    const selectedValue = questions.options[questions.selectedIndex].value
```
jessika-hage, line 269: Here you could just write const value = questions.value. A little shorter! Then you could also just put value, in the if statements instead of value: selectedValue :)

code/script.js
```diff
@@ -241,103 +243,152 @@ const generateBoard = () => {
...
-  // Show the correct alert message for different categories
-  if (group === 'accessories') {
-    if (keep) {
-      alert(
-        `Yes, the person wears ${attribute}! Keep all that wears ${attribute}`
-      )
+    /*declaring the different variable from the checkQuestion function*/
+    let attribute = currentQuestion.attribute
+    let value = currentQuestion.value
+    let group = currentQuestion.category
```
jessika-hage, line 328: Could use const here too

jessika-hage (approved):
You did a great job, Hannah! The code looks very clean and easy to understand. Just had a few comments, but nothing that is necessary to change I think. This was not an easy project but you did a really good job 👍 :D

code/script.js
jesinkey: OK, cool! I didn't know that.

code/script.js
KarolCislowski: You need to fix your indentation in your VScode settings. It should be set to `2 spaces`.

KarolCislowski (approved):
This is a very nice project 🌈 

PR #25, opened 2021-02-13
Project 5 - Mette

ettem:
The one thing I'm having problems understanding completely is why I had to change the value for the objects in index to make the javascript file work as I wished for. Were we supposed to do that or did I construct that need myself writing the javascript code? I still think it is a bit difficult to separate value, attribute and category for the objects, I keep mixing them up!


code/index.html
```diff
@@ -3,45 +3,61 @@
...
+    href="https://fonts.googleapis.com/css2?family=Comfortaa:wght@500&display=swap" 
+    rel="stylesheet"
+    />
   </head>
   <body>
     <aside class="question-section">
-      <button id="restart" class="outlined-button">RESTART</button>
-
+  
+      <button id="restart" class="buttonRestart">RESTART</button>
```
rubmaigut, line 21: I would recommend naming classes with snake-case. This is no affecting the code, but it's considered a standard way to name classes.

code/script.js
```diff
@@ -246,98 +249,170 @@ const generateBoard = () => {
 
 // Randomly select a person from the characters array and set as the value of the variable called secret
 const setSecret = () => {
-  secret = charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)]
+  let randomElement = Math.random() * charactersInPlay.length //note to self! for a randomised number rounded off to length of array (will be a decimal (float) and therefore math.floor after)
+  let randomElementInt = Math.floor(randomElement);
```
rubmaigut, line 253: Good form to change the initial logic, but knowing that the values of randomElement and randomElementInt are not being reassigned does two are more like const.

code/script.js
```diff
@@ -246,98 +249,170 @@ const generateBoard = () => {
...
+    }
   }
 }
 
 // This function should be invoked when you click on 'Find Out'.
 const checkQuestion = () => {
   // Compare the currentQuestion with the secret person.
   // See if we should keep or remove people based on that
+  let keepPerson;
+  let checking = currentQuestion.attribute;
```
rubmaigut, line 313: It's the same in this line, the value of checking is not being reassigned. In this case, is `const`

code/script.js
```diff
@@ -246,98 +249,170 @@ const generateBoard = () => {
...
   // Invoke a function to redraw the board with the remaining people.
+  generateBoard();
 }
 
 // when clicking guess, the player first have to confirm that they want to make a guess.
 const guess = (suspect) => {
   // store the interaction from the player in a variable.
   // remember the confirm() ?
-  // If the player wants to guess, invoke the checkMyGuess function.
+  let choice = confirm(`Do you want to guess ${suspect}?`);
```
rubmaigut, line 383: Change this as `const` as well.

rubmaigut (changes requested):

rubmaigut:
Good Job in this project, I think that with the change in the variable we can approve the PR

code/index.html
KarolCislowski: Yes, this is Technigo's naming convention so all CSS selectors should be written in kebab-case. 🍡

KarolCislowski (approved):
This is a very nice project 🌈 

PR #26, opened 2021-02-13
Project 5 - Guess Who - Erika A Porath

TheErry:
This is my finished game. The game is working as it should, even though I couldn't get some of the code we were shown (like shorthands) to work for some reason. But this code works :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,13 @@
 # Project Name
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+This weeks project was to create a Guess Who game by using Javascript, and foremost arrays and objects.
 
 ## The problem
 
 Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
 
-* Saw what my teammates hade done to start, or else I wouldn't have figured it out.
+I had trouble understanding how to start the project at first. The steps in the brief were helpful, but I still felt confused. It helped a bit when I saw how my team-mates hade done it in their projects. I coded along with the lesson on wednesday, and that helped a lot! But later on I started to get confused again, mainly because I swithed focus between the pre-written code, the lessons and the steps. Once again I got great help from my team-mates to solve it. The rest I managed to figure out by looking up info on Stack Overflow and other sites, and one time I looked up someone else project on GitHub to see how they solved the problem. It turned out that I was mostly on the correct path, and that most of my errors were typos or that I forgot to put an alert() in the correct place and such. If I had more time I would love to add an counter and a timer, and to show the result on the win-or-lose-page. I would also like to style it a bit more, and it would be fun to add my own family instead. But the code is working now, and I'm satisfied with that.
 
 ## View it live
 
-Have you deployed your project somewhere? Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+You can see my project here:
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ Start by briefly describing the assignment in a sentence or two. Keep it short a
 
 Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
 
+* Saw what my teammates hade done to start, or else I wouldn't have figured it out.
+
 ## View it live
 
 Have you deployed your project somewhere? Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
-# Project Name
+# Guess Who game project
 
 This weeks project was to create a Guess Who game by using Javascript, and foremost arrays and objects.
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
-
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,13 +1,13 @@
...
+I had trouble understanding how to start the project at first. The steps in the brief were helpful, but I still felt confused. It helped a bit when I saw how my team-mates hade done it in their projects. I coded along with the lesson on wednesday, and that helped a lot! But later on I started to get confused again, mainly because I swithed focus between the pre-written code, the lessons and the steps. Once again I got great help from my team-mates to solve it. The rest I managed to figure out by looking up info on Stack Overflow and other sites, and one time I looked up someone else project on GitHub to see how they solved the problem. It turned out that I was mostly on the correct path, and that most of my errors were typos or that I forgot to put an alert() in the correct place and such. If I had more time I would love to add an counter and a timer, and to show the result on the win-or-lose-page. I would also like to style it a bit more, and it would be fun to add my own family instead. Also sometimes the code doesn't work the way we were shown, and I would like to have time to figure out why. But the code is working now as it is, and I'm satisfied with that.
```
katliebig, line 7: Smart that you use several ways to get help, ask your team, checking StackOverflow and looking at other's solutions. Remember that the lectures mostly only show one approach. There's always several ways of doing something, and the most important thing is that you understand the approach you have chosen.

code/script.js
```diff
@@ -251,48 +255,73 @@ const setSecret = () => {
...
+  board.style.display = 'flex'
+  winOrLose.style.display = 'none'
+  questions.selectedIndex = defaultStatus
+  generateBoard()
+  setSecret()
 }
 
-// setting the currentQuestion object when you select something in the dropdown
+// Setting the currentQuestion object when you select something in the dropdown
+// ! It doesn't work with shorthands for some reason. I know I could use it here though !
```
katliebig, line 267: I haven't even tried the shorthand approach, so good job for trying! Maybe practice with an easier, less nested example so you know how the shorthand's working.

code/script.js
```diff
@@ -251,48 +255,73 @@ const setSecret = () => {
...
-  // Compare the currentQuestion with the secret person.
-  // See if we should keep or remove people based on that
-  // Then invoke filterCharacters
-}
+const checkQuestion = (selectQuestion) => {
+  let keep
 
-// It'll filter the characters array and redraw the game board.
+  if (currentQuestion.category === 'hair color') {
+      keep = currentQuestion.value === secret.hairColor
```
katliebig, line 304: This is really smart!

code/script.js
```diff
@@ -304,40 +333,74 @@ const filterCharacters = (keep) => {
...
 
 // If you confirm, this function is invoked
 const checkMyGuess = (suspect) => {
-  // 1. Check if the suspect is the same as the secret person's name
-  // 2. Set a Message to show in the win or lose section accordingly
-  // 3. Show the win or lose section
-  // 4. Hide the game board
-}
+  if (suspect === secret.name) {
+    winOrLoseText.innerHTML = `${suspect} is correct! You won!`
```
katliebig, line 390: Nice that you're re-using the suspect variable in the text shown to the user.

code/script.js
```diff
@@ -251,48 +255,73 @@ const setSecret = () => {
...
+      keep = secret.hat === true
+    }
+  } else {
+      keep = secret.smoker === true
+  }
+
+  filterCharacters(keep)
+} 
+
+// Filter the characters array and redraw the game board.
```
katliebig, line 320: Good job with the comments!

code/script.js
```diff
@@ -251,48 +255,73 @@ const setSecret = () => {
...
 const filterCharacters = (keep) => {
-  // Show the correct alert message for different categories
+  let group = currentQuestion.category
+  const attribute = currentQuestion.attribute
+
   if (group === 'accessories') {
```
katliebig, line 325: These nested ifs are a bit hard to read. I know that the project was structured this way to begin with (and in the beginning, I also had these nested ifs in my code), and as your project is working perfectly, it's good the way it is. 
In case you want to rewrite your code, I would like to share that someone on Slack posted a really nice solution on how to make this less messy. If you give the objects created inside selectQuestion another property, text, and store the innerText of the select menu options as the value of text (so that for eyecolor, it can say "brown eyes"), you can use the text property to fill in the alert text. And that way, you only need one alert for yes and one for no. If you want to, you can have a look at my solution. Not sure if I explained it well - let me know if you have questions about that.

code/script.js
```diff
@@ -251,48 +255,73 @@ const setSecret = () => {
...
+  }
+
+  filterCharacters(keep)
+} 
+
+// Filter the characters array and redraw the game board.
 const filterCharacters = (keep) => {
-  // Show the correct alert message for different categories
+  let group = currentQuestion.category
+  const attribute = currentQuestion.attribute
```
katliebig, line 323: I think these variable can both be declared using const instead of let.

katliebig (approved):
Really nice job! It's up and running as it should, and you can be proud of that! :)
I think your code is very clean and consistent, nice indentation. I did not get stressed out when I looked at it, and that surely is a good sign!

code/script.js
```diff
@@ -251,48 +255,73 @@ const setSecret = () => {
...
-      // 👆 add the value from the input here
+      value: value,
       category: category,
     }
   } else if (category === 'eye color') {
-    // Set this up your self
+    currentQuestion = {
+      attribute: 'eyeColor',
+      value: value,
+      category: category,
```
KarolCislowski, line 282: This object could be written shorter:
```
{
      attribute: 'eyeColor',
      value,
      category
}
```

code/script.js
KarolCislowski: `switch` would be a better way to write this cleaner 🤓

KarolCislowski (approved):
This is a very nice project 🌈 

PR #35, opened 2021-02-14
Project Guess Who Anna Lenksjö

annalenksjo:


code/script.js
```diff
@@ -253,91 +257,144 @@ const setSecret = () => {
...
+        charactersInPlay = charactersInPlay.filter((person) => person[currentQuestion.attribute] === currentQuestion.value);
+      } else {
+        alert(
+          `No, the person does not have ${attribute} eyes! Remove all that have ${attribute} eyes`
+        )
+        charactersInPlay = charactersInPlay.filter((person) => person[currentQuestion.attribute] !== currentQuestion.value);
+      }
+    }
+    generateBoard ()
 }
```
TobiasHagberg, line 364: I offer this comment based on that we chose to solve the problem given in different ways. 

First, I applaud the functionality! 👏

My thought process taking on this problem...
I really wanted to make use of the code provided and got stuck wrapping my head around the filtering and how the global variables were to be implemented throughout the game.

They are all set into play i this section and we know that the global variables are set to: let secret, currentQuestion, charactersInPlay.

- The user makes a simple choice among the questions and this toghether with clicking the find out button  triggers a series of functions.

    1.  We listen for the choice/change with the event listener - selectQuestion
    2. selectQuestion _charges_ the global varible currentQuestion with a value after checking the users catergory choice and validating it against the _secret_
    3. It then goes into the filterCharacters statement and prompting alerts according to selection and then the filter method creates a "copy" of the original board and _charactersInPlay_  on wich it bases the regenerated board.
   
The implied suggestion of use making use of group really throw me off and I just had to change it into something we used earlier in the code, in this case filterCharacters. I don't know if this is really the case but I think this is a reason why I was able to seperate the if statement frpm the filtering and having them execute on different lines instead of on every statement?

This is my code fore reference and I hope this comment makes sense, it atleast helped me clear my own thoughts :)

`const filterCharacters = (keep) => {
  if (currentQuestion.category === 'accessories') {
    if (keep) {
      alert(`Yes, the person wears ${currentQuestion.attribute}! Keep all that wears ${currentQuestion.attribute}`)
    } else {
      alert(`No, the person doesn't wear ${currentQuestion.attribute}! Remove all that wears ${currentQuestion.attribute}`)
    }
  } else if (currentQuestion.category === 'eye color') {
    if (keep) {
      alert(`Yes, the person has ${currentQuestion.value} eyes! Keep all that has ${currentQuestion.value} eyes`)
    } else {
      alert(`No, the person doesn't have ${currentQuestion.value} eyes! Remove all that has ${currentQuestion.value} eyes`)
    }  
  } else if (currentQuestion.category === 'hair color') {
    if (keep) {
      alert(`Yes, the person has ${currentQuestion.value} hair! Keep all that has ${currentQuestion.value} hair`)
    } else {
      alert(`No, the person doesn't have ${currentQuestion.value} hair! Remove all that has ${currentQuestion.value} hair`)
    }  
  } else if (currentQuestion.category === 'other') {
    if (keep) {
      alert(`Yes the person smokes! Keep all the smokers`)
    } else {
      alert(`No the person doesn't smoke! Remove all the smokers`)
    }
  }

  if (keep) {
    charactersInPlay = charactersInPlay.filter((person) => person[currentQuestion.attribute] === currentQuestion.value)
  } else {
    charactersInPlay = charactersInPlay.filter((person) => person[currentQuestion.attribute] !== currentQuestion.value)
  }
  generateBoard()
}`

code/script.js
```diff
@@ -253,91 +257,144 @@ const setSecret = () => {
...
 
 // when clicking guess, the player first have to confirm that they want to make a guess.
 const guess = (suspect) => {
-  // store the interaction from the player in a variable.
-  // remember the confirm() ?
-  // If the player wants to guess, invoke the checkMyGuess function.
-}
```
TobiasHagberg, line 329: Can't see that this function is ever used? Delete?

code/script.js
```diff
@@ -253,91 +257,144 @@ const setSecret = () => {
...
+  if (secret.name === suspect) {
+    let winSound = document.getElementById('winSound')
+    winSound.play()
+   winOrLoseText.innerHTML=`Congratulations, you guessed right!!`
+  } else {
+    let loseSound = document.getElementById('loseSound')
+    loseSound.play()
+    winOrLoseText.innerHTML=`Oh no, your guess was wrong :-(`
+  }
 }
```
TobiasHagberg, line 392: Nice touch with both inner.HTML and sounds!

TobiasHagberg (commented):
Hej Anna,
Tack för att jag fick ta del av din kod, det har återigen visat att det finns många möjliga lösningar på samma problem. Jag har lärt mig mycket genom att försöka analysera och jämföra med min egen kod. Mina kommentarer grundar sig i det och jag hoppas jag lyckats förmedla dem på ett bra sätt

README.md
```diff
@@ -1,13 +1,12 @@
...
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+Project 'Guess Who' - a project that was made during the Technigo Bootcamp spring 2021. It is based on the classic board game.
 
 ## The problem
+I started by reading up on the techniques I needed to use, for example filtering method. Technigo provided a sort of step by step structure to start the project so I followed that, and then I worked with the functions in the same order as the game goes. 
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+If I had more time I would work with 'drying' the code and making it more efficient. I would also add my own styling.
```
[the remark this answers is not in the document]
KarolCislowski: `Don't Repeat Yourself` so drying is to make your code shorter and more efficient. The more professional term for this is `refactoring`.

There is also `WET - We Enjoy Typing` as opposed to the `DRY` methodology.

code/script.js
KarolCislowski: Yes, remember to always clean your code from all console.log()s and commented out code blocks and tips left for you by teachers 🗡️ ✂️

code/script.js
```diff
@@ -253,91 +257,144 @@ const setSecret = () => {
...
-      // 👆 add the value from the input here
+      value: optionValue,
       category: category,
     }
   } else if (category === 'eye color') {
-    // Set this up your self
+    currentQuestion = {
+    attribute: 'eyeColor',
+    value: optionValue,
+    category: category,
```
KarolCislowski, line 282: remember that you do not need to add value when you want to pass the variable with the same name as the key. 
`{category: category}` can be written as `{category}` 👍🏻 And when you have a bigger project you can use saved time to watch some Netflix 🙈

KarolCislowski (approved):
This is a very nice project 🌈 

PR #37, opened 2021-02-14
Guess Who

CarolineF3:


code/confetti.js
```diff
@@ -0,0 +1,213 @@
+var confetti = {
```
darcywinterhill, line 1: Fun function!

code/script.js
```diff
@@ -262,82 +270,148 @@ const selectQuestion = () => {
...
+  if (currentQuestion.value === secret[currentQuestion.attribute]) {
+    keep = true
+  } else {
+    keep = false
+  }
+  filterCharacters(keep)
 }
 
 // It'll filter the characters array and redraw the game board.
 const filterCharacters = (keep) => {
```
darcywinterhill, line 319: It's easy to understand where all code (group, attribute, value) is coming from and how it fits in the filter function.

code/script.js
```diff
@@ -262,82 +270,148 @@ const selectQuestion = () => {
...
 }
 
 // If you confirm, this function is invoked
 const checkMyGuess = (suspect) => {
-  // 1. Check if the suspect is the same as the secret person's name
-  // 2. Set a Message to show in the win or lose section accordingly
-  // 3. Show the win or lose section
-  // 4. Hide the game board
+
+  let message = `Sorry, you lost! The secret person was ${secret.name}.`
```
darcywinterhill, line 391: If I understand it right the variable 'message' has the value of 'Sorry, you lost!' from the start? If it is so, then a suggestion to make it more clear could be to have it empty and then get its value in the if and else depending on whether you win or lose.

code/index.html
```diff
@@ -1,64 +1,63 @@
...
+        <option value="hat">hat</option>
+
+      </optgroup>
+      <optgroup label="other">
+        <option value="smoker">a smoking habit</option>
+      </optgroup>
+    </select>
+
+    <button id="filter" class="filled-button">FIND OUT</button>
+    <div class="guess-counter" id="guess-counter" </div> </aside> <section class="board-wrapper">
```
darcywinterhill, line 52: I didn't know about the </aside> tag and googled it to find out what it does haha. 
Have been trying to figure out what it does specifically in your code, but it could just be me not seeing it!

code/script.js
```diff
@@ -254,6 +255,13 @@ const start = () => {
   // Here we're setting charactersInPlay array to be all the characters to start with
   charactersInPlay = CHARACTERS
   // What else should happen when we start the game?
+  generateBoard()
+  setSecret()
+  guesses = 0
+  guessCounter.innerHTML = `<p>Number of guesses: ${guesses}</p>`
```
darcywinterhill, line 261: I like this function to keep the player notified how many guesses the've made!

darcywinterhill (approved):
Clean code that is easy to read and understand! Fun extra features and functions beyond the minimum requirements :)

code/confetti.js
KarolCislowski: According to Technigo's code guide, usage of `var` is not allowed. please use `let` or `const` 😸

KarolCislowski (approved):
This is a very nice project 🌈 

Remember that you can always come back to it when you have some spare time and add more black level functionalities to make it an even greater portfolio asset! 🦄

PR #38, opened 2021-02-14
Project 5, Guess Who / Anna-Kaisa Honkonen

honkoan:
https://guess-who-honkoan.netlify.app/

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -304,12 +304,11 @@ const selectQuestion = () => {
 const checkQuestion = () => {
   let keep
   let secretValue = secret[currentQuestion.attribute]
-  if (currentQuestion === secretValue) {
+  if (currentQuestion.value === secretValue) {
     keep = true
   } else {
     keep = false
   }
-  console.log (keep)
 
   filterCharacters(keep)
   // Compare the currentQuestion with the secret person.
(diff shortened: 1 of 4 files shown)


code/script.js
```diff
@@ -247,97 +252,175 @@ const generateBoard = () => {
...
     currentQuestion = {
-      //attribute: ,
+      attribute: value, // do something here!!!
       // 👆 this is the property of the booleans such as smoke, glasses and hat. add the value from the input here
       value: true, // we're asking if this person wears a hat for exaple, so always true in the question.
-      category: category,
+      category
     }
   } else if (category === 'other') {
+    currentQuestion = {
```
Irina-web-dev, line 298: You are missing property "attribute: actualValue " here. This might be the reason why smoking filtering is not working in the game.

code/script.js
```diff
@@ -247,97 +252,175 @@ const generateBoard = () => {
...
       alert(
-        `No, the person doesn't wear ${attribute}! Remove all that wears ${attribute}`
+        `No, the person doesn't wear ${[currentQuestion.attribute]}! Remove all that wears ${[currentQuestion.attribute]}`
       )
+      charactersInPlay = charactersInPlay.filter((person) => person[currentQuestion.attribute] !== currentQuestion.value) 
     }
-  } else if (group === 'other') {
-    // Similar to the one above
-  } else {
+  } else if (category === 'other') {
```
Irina-web-dev, line 340: Smoking filtering is not working. One reason is that you haven´t defined "attribute: actualValue " property (see my comment in line 298). And here similar approach as for other categories´ alert messages should be used: 
 
else if (category === 'other') {
 if (keep) {
      alert(
        `Yes, the person is a ${currentQuestion.attribute}! Keep all persons who are ${attribute}`
 )
    } else {
      alert(
        `No, the person is not a ${currentQuestion.attribute}! Remove all persons who are not ${attribute}`
      )
    }

Irina-web-dev (changes requested):
Awesome project and great job. It was easy to read your code, everything is clean and clear. I can see that you have applied knowledge gained through the code sessions, for example, you have used objects shorthand in your project.  It is good to start applying this knowledge and skills from the very beginning.

Below I have provided some tips about using comments and console.logs.

And I have noticed that "smoker" filtering is not working in the game. Due to that I would ask you to make changes before approval.  I have also provided advice for this problem.

code/script.js
```diff
@@ -251,93 +256,146 @@ const setSecret = () => {
...
+  } else {
+    keep = false
+  }
+
+  filterCharacters(keep)
 }
 
-// It'll filter the characters array and redraw the game board.
+// Filter the characters array and redraw the game board.
 const filterCharacters = (keep) => {
```
KarolCislowski, line 310: You creating the `keep` here, so this function does not use the global one. There is no use in storing extra global copy then 🤓

KarolCislowski (approved):
This is a very nice project 🌈 

It's great that you followed your friends tips. It is very important to act after receiving the review 🥳

PR #42, opened 2021-02-14
Project guess who - Lars Söderman

lars-soderman:
Ok here is a pull request for a very vanilla guess who!


code/script.js
```diff
@@ -244,100 +246,115 @@ const generateBoard = () => {
...
-  // This variable stores what option group (category) the question belongs to.
-  // We also need a variable that stores the actual value of the question we've selected.
-
+  const value = questions.options[questions.selectedIndex].value
   if (category === 'hair color') {
     currentQuestion = {
       attribute: 'hairColor',
-      // value: ,
-      // 👆 add the value from the input here
+      value: value,
```
Sofarz, line 270: Here you could use the object property shorthand: "value," and "category," since the key and the value are the same.

code/script.js
```diff
@@ -244,100 +246,115 @@ const generateBoard = () => {
...
     }
   } else if (category === 'eye color') {
-    // Set this up your self
-  } else if (category === 'accessories') {
     currentQuestion = {
-      //attribute: ,
-      // 👆 this is the property of the booleans such as smoke, glasses and hat. add the value from the input here
-      value: true, // we're asking if this person wears a hat for exaple, so always true in the question.
+      attribute: 'eyeColor',
+      value: value,
```
Sofarz, line 276: Same goes here.

code/script.js
```diff
@@ -244,100 +246,115 @@ const generateBoard = () => {
...
+      category: category,
+    }
+    
+  } else if (category === 'accessories' || category === 'other') {
+    currentQuestion = {
+      attribute: value,
+      value: true, 
       category: category,
```
Sofarz, line 284: Same here.

code/script.js
```diff
@@ -244,100 +246,115 @@ const generateBoard = () => {
...
-    } else {
-      alert(
-        `No, the person doesn't wear ${attribute}! Remove all that wears ${attribute}`
-      )
-    }
-  } else if (group === 'other') {
-    // Similar to the one above
+  // Alerts...
+  if (currentQuestion.category === 'accessories') {
+    keep ?
```
Sofarz, line 302: Maybe it would be easier to read the code if you would have continued with if-else statements instead of ternary operators here?

code/script.js
```diff
@@ -244,100 +246,115 @@ const generateBoard = () => {
...
 }
 
-// If you confirm, this function is invoked
 const checkMyGuess = (suspect) => {
-  // 1. Check if the suspect is the same as the secret person's name
-  // 2. Set a Message to show in the win or lose section accordingly
-  // 3. Show the win or lose section
-  // 4. Hide the game board
+  if (suspect === secret.name) {
+    board.innerHTML = ''
```
Sofarz, line 343: Is this line necessary?

code/script.js
```diff
@@ -244,100 +246,115 @@ const generateBoard = () => {
...
+    winOrLose.style.display = 'block'
+    winOrLoseText.innerHTML = 'You won - yay!'
+    
+  } else {
+    board.innerHTML = ''
```
Sofarz, line 348: Is this line necessary?

Sofarz (approved):
Great job on your Guess who game! 💯 👍 The game works just as it should and I have only added a few comments on minor code improvements and suggestions to avoid unnecessary code and to make it more readable :). Keep up the good work!

code/script.js
lars-soderman: Hey Sofia, thanks for a great code review! My idea for line 343 (and 348) was to empty the game board before displaying the win or lose-message!

code/script.js
KarolCislowski: or use `switch`?

KarolCislowski (approved):
This is a very nice project 🌈 

PR #45, opened 2021-02-14
Guess Who project

sarianne-peltola:
Hi,
I finished with red level requirements.

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -254,6 +254,7 @@ const start = () => {
   // Here we're setting charactersInPlay array to be all the characters to start with
   charactersInPlay = CHARACTERS
   // What else should happen when we start the game?
+  generateBoard();
 }
 
 // setting the currentQuestion object when you select something in the dropdown
--- a/code/script.js
+++ b/code/script.js
@@ -226,7 +226,7 @@ const CHARACTERS = [
 ]
 
 // Global variables
-let secret, currentQuestion, charactersInPlay
+let secret, currentQuestion, charactersInPlay, keep
 
 // Draw the game board
 const generateBoard = () => {
@@ -301,12 +301,19 @@ const selectQuestion = () => {
 
 // This function should be invoked when you click on 'Find Out'.
 const checkQuestion = () => {
-  console.log("is the button working?")
   selectQuestion()
-  console.log(currentQuestion)
+  //console.log(currentQuestion)
+  if (currentQuestion["value"] === secret[currentQuestion["attribute"]]) {
+    console.log(true)
+    let keep = true
+  } else {
+    console.log(false)
+    let keep = false
+  }
   // Compare the currentQuestion with the secret person.
   // See if we should keep or remove people based on that
   // Then invoke filterCharacters
+ // filterCharacters(keep)
 }
 
 // It'll filter the characters array and redraw the game board.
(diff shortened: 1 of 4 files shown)


code/script.js
```diff
@@ -247,97 +300,186 @@ const generateBoard = () => {
...
 }
 
-// It'll filter the characters array and redraw the game board.
+// This will filter the characters array and redraw the game board.
 const filterCharacters = (keep) => {
-  // Show the correct alert message for different categories
-  if (group === 'accessories') {
-    if (keep) {
+  if (currentQuestion.category === 'accessories') {
+    if (keep === true) {
```
takrami, line 386: It can be skipped. To check boolean variable in a if condition there is no need to put `=== true`. The suggestion and the original code are identical and behave the same in this specific case. 

```suggestion
    if (keep) {
```

code/script.js
```diff
@@ -247,97 +300,186 @@ const generateBoard = () => {
...
+  generateBoard()
 }
 
 // when clicking guess, the player first have to confirm that they want to make a guess.
 const guess = (suspect) => {
-  // store the interaction from the player in a variable.
-  // remember the confirm() ?
-  // If the player wants to guess, invoke the checkMyGuess function.
-}
+  let confirmation = confirm("Are you sure you want to make a guess?");
```
takrami, line 463: There is not another assignment for this variable in this scope.

```suggestion
  const confirmation = confirm("Are you sure you want to make a guess?");
```

code/script.js
```diff
@@ -247,97 +300,186 @@ const generateBoard = () => {
...
-  // 3. Show the win or lose section
-  // 4. Hide the game board
+  if (suspect === secret.name) {
+    winOrLoseText.innerHTML = "You won!!! Congrats!"
+    winOrLose.style.display = "flex";
+  } else {
+    winOrLose.style.display = "flex";
+    winOrLoseText.innerHTML = "Game over... Better luck next time."
+  }
+
```
takrami, line 477: In either scenarios the code is applying this: `winOrLose.style.display = "flex";`. It can be moved at the end.
```suggestion
  if (suspect === secret.name) {
    winOrLoseText.innerHTML = "You won!!! Congrats!"
  } else {
    winOrLoseText.innerHTML = "Game over... Better luck next time."
  }
  winOrLose.style.display = "flex";
```

takrami (approved):
Great job :tada:  I really enjoyed reading your code. So nice and clean :+1: I wrote down a few tiny comments which might help the code quality. I suggest to use https://prettier.io/ . It made my life easier while coding :smile:  might help you as well. Basically it formats the code indentation and missing semicolons and removes unnecessary blank lines and stuff like that. LGTM :100:

code/script.js
```diff
@@ -247,97 +300,186 @@ const generateBoard = () => {
...
+  filterCharacters(keep)
+  count += 1;
+  console.log(count)
+  countQuestion.innerHTML = `${count}`
 }
 
-// It'll filter the characters array and redraw the game board.
+// This will filter the characters array and redraw the game board.
 const filterCharacters = (keep) => {
```
KarolCislowski, line 384: You are creating the internal `keep` variable. This function does not use the global `keep` so it is no need to create it.
You always should avoid creating global variables.

KarolCislowski (approved):
This is a very nice project 🌈 

Remember that you can always come back to it when you have some spare time and add more black level functionalities to make it an even greater portfolio asset! 🦄

PR #50, opened 2021-02-14
Project guess who 

liaback-lowstett:


code/style.css
```diff
@@ -61,7 +64,7 @@ select {
 
 /****** CARD ******/
 .card {
-  width: 135px;
+  width: 120px; /* changed width to fit in screen-view whitout scrolling*/
```
lars-soderman, line 67: 🚀

lars-soderman (commented):
Hey Felicia 👋  Really great job with the guess who-game! Everything works like it should, so it is all good. I put in some thoughts and comments in the code. 
Rock on! 💥

code/script.js
```diff
@@ -247,53 +250,86 @@ const generateBoard = () => {
...
+  // Invoke filterCharacters
+  filterCharacters(keep)
 }
 
-// It'll filter the characters array and redraw the game board.
+// Filter the characters array and redraw the game board.
 const filterCharacters = (keep) => {
+  let category = currentQuestion.category
+  let attribute = currentQuestion.attribute
+  let value =currentQuestion.value
```
KarolCislowski, line 328: `const { category, attribute, value } = currentQuestion` would do the magic here 🧙🏻

KarolCislowski (approved):
This is a very nice project 🌈 

PR #53, opened 2021-02-14
Guess-who-game-Sandra-Wallen

sandra-wallen:


code/script.js
```diff
@@ -249,95 +251,111 @@ const setSecret = () => {
...
-// This function to start (and restart) the game
+// Function to start the game
 const start = () => {
-  // Here we're setting charactersInPlay array to be all the characters to start with
   charactersInPlay = CHARACTERS
-  // What else should happen when we start the game?
+  generateBoard()
+  setSecret()
+
+  console.log(secret)
```
ElleFrankenberg, line 260: I don't think you need to keep the console.log in your code.

code/style.css
```diff
@@ -159,6 +159,10 @@ button {
   height: 100vh;
 }
 
+.show-win-or-lose-section {
+  display: flex;
+}
+
```
ElleFrankenberg, line 165: I don´t know wish way is the best, but I (or rather Isabella how helped me) put this directtielly in the script instead. Like this:

const start = () => { winOrLose.style.display = 'none' board.style.display ='flex'

ElleFrankenberg (approved):
Hey Sandra😀
As you know my JavaScript knowledge is a bit limited.... With that said I think your code is looking good, the game works and that the blue requirements are fulfilled.
Good job! And thank you again for all the help last week!
Best,
Elle

code/script.js
KarolCislowski: Yes, remember to always clean your code from all console.log()s and commented out code blocks and tips left for you by teachers 🗡️ ✂️

code/script.js
```diff
@@ -249,95 +251,111 @@ const setSecret = () => {
...
+      value: value,
       category: category,
+      alertText: `${value} hair`
     }
   } else if (category === 'eye color') {
-    // Set this up your self
+    currentQuestion = {
+      attribute: 'eyeColor', 
+      value: value,
+      category: category,
```
KarolCislowski, line 279: remember that you do not need to specify the `value` when both sides of the `key: value` pair are the same 🤓

KarolCislowski (approved):
This is a very nice project 🌈 

PR #65, opened 2021-09-17
Guess who - Harry Bäcklin

Groovedharry:


code/script.js
```diff
@@ -204,6 +204,9 @@ const CHARACTERS = [
 let secret
 let currentQuestion
 let charactersInPlay
+let category
+let value
+let keep = true;
```
isomoth, line 209: I'm curious about the use of semicolons in this section. Did you have something specific in mind when adding a semicolon in line 209, but not to the rest of the global variables? 

This is a really minor detail, nothing that affects the code performance or such :) But for style consistency, I think there could be two options: 

```
let secret
let currentQuestion
let charactersInPlay
let category
let value
let keep = true

```

Or (I tend to use this one, but it's really a matter of taste): 

```
let secret, 
    currentQuestion, 
    charactersInPlay, 
    category, 
    value, 
    keep = true;
```

Maybe using the Prettier extension on VS Code helps, too?

code/script.js
```diff
@@ -232,94 +235,175 @@ const start = () => {
   // Here we're setting charactersInPlay array to be all the characters to start with
   charactersInPlay = CHARACTERS
   // What else should happen when we start the game?
-}
+  generateBoard();
+  setSecret();
+  // console.log(secret)
```
isomoth, line 240: If my memory doesn't fail me, I think Maks recommended to get rid of the console.logs and unnecessary comments once you're ready to hand in your code. It's definitely not a problem in here, but it could get cluttered quickly if you escalated your project to a bigger document :)

code/script.js
```diff
@@ -232,94 +235,175 @@ const start = () => {
...
   if (category === 'hair' || category === 'eyes') {
+      if (secret.hair === value || secret.eyes === value) {
+          filterCharacters(keep)
+      } else {
+          filterCharacters()
+      }
 
   } else if (category === 'accessories' || category === 'other') {
-
+      if (secret.accessories[0] === value || secret.other[0] === value) {
```
isomoth, line 275: I would also think in terms of scalability here. Right now, hard-coding the index of the accessories array isn't a problem because there is only one index anyway. But if you wanted to add more items, you'd need to use a variable instead of an index number. 

You could do something similar to line 289, where you used `person[category]` instead of a hard-coded index.

code/script.js
```diff
@@ -232,94 +235,175 @@ const start = () => {
...
   // store the interaction from the player in a variable.
   // remember the confirm() ?
   // If the player wants to guess, invoke the checkMyGuess function.
+  let confirmVar = confirm("Are you sure?")
+  // console.log(personToConfirm +' test')
+  if (confirmVar == true) {
+    // console.log('something3')
+    checkMyGuess(personToConfirm)
+  } else {
+    stop
```
isomoth, line 369: Interesting, I haven't seen this before. Is this the same as `window.stop();`?

code/script.js
```diff
@@ -232,94 +235,175 @@ const start = () => {
...
+  }
+}
+function reloadF(){
+  window.location.reload()
```
isomoth, line 373: Good idea! I hadn't thought about using that method.

code/script.js
```diff
@@ -232,94 +235,175 @@ const start = () => {
...
 }
 
 // If you confirm, this function is invoked
 const checkMyGuess = (personToCheck) => {
+  let txt = ''
+  // console.log('something2')
+  let winOrLoseTextElement = document.querySelector('#winOrLoseText');
+  let winOrloseWrapperElement = document.querySelector('.win-or-lose-wrapper')
```
isomoth, line 381: Interesting use of the querySelectors here. Was there any particular reason you decided to use querySelectors instead of `getElementByID` and `getElementsByClassName` as you did with the variables at the start? 

For the purpose of these variables, I went with `getElementByID`. They tend to have a [faster performance](https://gomakethings.com/javascript-selector-performance/). Not suuuper fast, but still :) 

 If you needed to escalate to [a more elaborate rule](https://careerkarma.com/blog/javascript-queryselector-vs-getelementbyid/) to select elements further down the road, then yeah, I'd go for querySelector.

isomoth (approved):
I had an easy time following your code as it was, no need for extra explanations or such. Good job! 

I think the important thing this week was to reinforce some JavaScript concepts and understanding a bit further. It shows that you do! The tutorial watching sessions are paying off 👍 

You're an independent learner, going beyond the course material. This is evident from the things you're trying in your code. It's a fantastic trait to have as a developer. Keep it up!

PR #73, opened 2021-09-17
Darya - Guess Who? 

DALA746:


code/confetti.js
```diff
@@ -0,0 +1,219 @@
+
+// THIS CODE I DONT QUIT UNDERSTAND BUT IT MAKES CONFETTI RAIN :)
+
+
```
Jsfrulle, line 4: Love that you use this effect 👯

code/confetti.js
```diff
@@ -0,0 +1,219 @@
...
+		particle.tiltAngleIncrement = Math.random() * 0.07 + 0.05;
+		particle.tiltAngle = Math.random() * Math.PI;
+		return particle;
+	}
+
+	function toggleConfettiPause() {
+		if (pause)
+			resumeConfetti();
+		else
+			pauseConfetti();
```
Jsfrulle, line 60: if you want to pause the confetti can you call pauseConfetti()?

code/script.js
```diff
@@ -1,7 +1,24 @@
...
+const playAgainButton = document.getElementById('playAgain')
+const startButton = document.getElementById('start')
+const filter = document.getElementById('filter')
+const startWindow = document.getElementById('start-window')
+
+const startAudio = document.getElementById('audio')
+const winningAudio = document.getElementById('audio-winning')
+const lostAudio = document.getElementById('audio-lost')
+
+const modal = document.getElementById("myModal");
```
Jsfrulle, line 13: maybe use a more describing name of this variable

code/script.js
```diff
@@ -221,105 +238,171 @@ const generateBoard = () => {
...
-
-    for accessories and other
-      charactersInPlay = charactersInPlay.filter((person) => person[category].includes(value))
-      or
-      charactersInPlay = charactersInPlay.filter((person) => !person[category].includes(value))
-  */
-
-  // Invoke a function to redraw the board with the remaining people.
+  const { category, value } = currentQuestion // taking value from category and value
+  modalWindow(category, value, keep) // calling this function and passing 3 paramenters, this function filter and showing message to the user
```
Jsfrulle, line 348: smart! :D

code/style.css
```diff
@@ -145,6 +184,31 @@ button {
...
+.start-div {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  max-width: 700px;
+  text-align: center;
+}
+
+
```
Jsfrulle, line 211: Dayra You did the game like I wish I had done mine SO IMPRESSIVE!!! I love the confetti, starter window, and custom popup!!!!! the only thing I think you could think about is to name the variables/tags to more describing - like starter window (if not the modal window is the right description for this variable and I just don't know that :) ) Good work!!! - -[ YES]  Am I able to understand the code easily?
- [NO ]  Does the code have lots of duplicates? Could it be tidied up?
- [ NO]  Does it work or does anything look broken?
- [YES]  Does it follow the general and blue requirements?

code/confetti.js
DALA746: Thanks Jessica! Yeah, really nice effect, but a lot of code that I don't understand :)

code/confetti.js
DALA746: Yes, I think so! It sounds logical. I will try to manipulate with that code :)

code/script.js
DALA746: I must be better to name classes, id and variables :) but sometimes it is hard. Thanks I will think of that in my future projects!

code/script.js
DALA746: Those functions can be a struggle, but when they work,  your code can be so much shorter and more easy to read :)

code/style.css
DALA746: Thank you Jessica for your code review! :) Really great inputs that I will apply for my next projects :)

PR #76, opened 2021-09-17
project-guess-who Jessica

Jsfrulle:


script.js
```diff
@@ -0,0 +1,446 @@
...
+		charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)];
+};
+
+
+const name = () =>{
+
+	let addName = prompt(`Hi Gamer, what is your name?` )
+	addName.value=""
+	nameOfGamer.innerText = `Welcome ${addName}`
+
```
AschwinSiaila, line 265: Very nice to ask the name.

script.js
```diff
@@ -0,0 +1,446 @@
...
+		`So you think ${suspected} is the suspected person? (yes or no) `
+	);
+
+	if (yourGuess === "yes") {
+		board.style.display = "none";
+		winOrLose.style.display = "flex";
+		if (secret.name === suspected) {
+			alert(`you win`);
+		} else if (secret.name !== yourGuess) {
+			alert(`you lose! This time the "Guess-Who-person" was ${secret.name}`);
```
AschwinSiaila, line 412: I was wondering if you can put the value of the name asking in the beginning of the game you can use here with ${ }

AschwinSiaila (commented):
This game looks good and has nice features. I like the counting of the guessing's and the asking of the name in the beginning. I was wondering if you can use the value of the name into the answers with ${ } 
		`So you think ${suspected} is the suspected person? (yes or no) `
			alert(`you win`);
			alert(`you lose! This time the "Guess-Who-person" was ${secret.name}`);
The game works, the audio works and it looks good.

script.js
Jsfrulle: Hi, I tried that but dint make it work.. do you have any solution to make that work? :)

to inser the ${addName} not working  


Tacks for your feedback!

PR #85, opened 2021-09-17
Guess who? by Anna Thunberg

annathunberg:


code/index.html
```diff
@@ -3,12 +3,13 @@
...
+    <title>Guess who?</title>
     <link rel="stylesheet" href="style.css" />
     <link
       href="https://fonts.googleapis.com/css2?family=Montserrat:ital,wght@0,100;0,200;0,300;0,400;0,500;0,600;0,700;0,800;0,900;1,100;1,200;1,300;1,400;1,500;1,600;1,700;1,800;1,900&display=swap"
       rel="stylesheet"
     />
+    <link rel="shortcut icon" href="/images/jazebelle.svg" type="image/x-icon">
   </head>
   <body>
     <aside class="question-section">
```
Zancotti, line 15: Never used this tag, interesting!

code/index.html
```diff
@@ -20,24 +21,34 @@
         alt="Guess Who"
       />
 
-      <h1>Does the person have</h1>
+      <h1>Does the person have...</h1>
 
       <select id="questions">
         <optgroup label="hair">
```
Zancotti, line 27: These sections are very nicely arranged, easy to overview!

code/script.js
```diff
@@ -222,104 +226,195 @@ const generateBoard = () => {
...
 // Randomly select a person from the characters array and set as the value of the variable called secret
 const setSecret = () => {
-  secret = charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)]
+ secret  = charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)]
 }
 
+
 // This function to start (and restart) the game
 const start = () => {
+  document.getElementById("winOrLose").style.display = "none";
```
Zancotti, line 238: Nice that you got this to work! (I had struggles with that in my code)

code/script.js
```diff
@@ -222,104 +226,195 @@ const generateBoard = () => {
...
+  generateBoard()
+  
+  // setSecret selects a secret character randomly from the CHARACTERS array
+  setSecret() 
+
+  // setting the currentQuestion object when you select something in the dropdown
+  selectQuestion()
 }
 
 // setting the currentQuestion object when you select something in the dropdown
```
Zancotti, line 253: I would suggest on trying to keep the same line spaces between everything. It makes the code a bit easier to read and looks very nice aswell!

code/script.js
```diff
@@ -222,104 +226,195 @@ const generateBoard = () => {
...
 const checkQuestion = () => {
-  const { category, value } = currentQuestion
+  const { category, value } = currentQuestion;
+
 
-  // Compare the currentQuestion details with the secret person details in a different manner based on category (hair/eyes or accessories/others).
-  // See if we should keep or remove people based on that
-  // Then invoke filterCharacters
   if (category === 'hair' || category === 'eyes') {
+    if (secret[category] === value) {
```
Zancotti, line 280: Nice solution for the filters! looks really clean =)

code/script.js
```diff
@@ -222,104 +226,195 @@ const generateBoard = () => {
...
 const guess = (personToConfirm) => {
-  // store the interaction from the player in a variable.
-  // remember the confirm() ?
-  // If the player wants to guess, invoke the checkMyGuess function.
+
+const userConfirmation = confirm("You sure? Please confirm!")
+if (userConfirmation == true) {
+   // If the player wants to guess, invoke the checkMyGuess function.
+  checkMyGuess(personToConfirm)
+} else {
```
Zancotti, line 375: I would remove the else, since the else statement is empty i dont think it has a function.

code/script.js
```diff
@@ -222,104 +226,195 @@ const generateBoard = () => {
...
+const checkMyGuess = (personToConfirm) => {
+  if(secret.name === personToConfirm) {
+    // Shows the win or lose section
+    winOrLose.style.display = "block"
+    // Hides the game board
+    board.style.display = "none"
+    // Message to show in the win section
+    winOrLoseText.innerHTML =
+    `<h1>Yaaayy you won!!</h1>`
+  } else {
```
Zancotti, line 392: On my cellphone part of the guess who logo is outside of the picture, i would put some padding above!

code/script.js
```diff
@@ -222,104 +226,195 @@ const generateBoard = () => {
...
+    board.style.display = "none"
+    // Message to show in the lose section
+    winOrLoseText.innerHTML =
+    `<h1>Sorry, you lost.</h1>`
+  }
+}
+
+// This function resets the board and hides winOrLose
+
+const playAgain = () => {
```
Zancotti, line 405: This function works really well, but as a suggestion another option would be to just put that information in line 406 and 407 in to the start function so that you dont need to have this function if you wanna save some space. Then you can just get to start from the click and it will automatically sort itself =)

Zancotti (commented):
Very well done Anna, It was a really tough assignment and you have done a great job in solving the game! You should be proud of yourself. I know how hard it is to feel lost in the code. 

Pat yourself on the back, and be proud! =)

code/index.html
annathunberg: I actually don't know why its called shortcut icon, the thing I usually use is called a favicon, but it seems to work either way! Nice touch to add a small icon in this way :)

code/script.js
annathunberg: Thank you for pointing this out! I usually make extra line spaces as I think its easier to read that way, but you are totally right that it's not good practice in a finished project.

code/script.js
annathunberg: Well thank YOU for helping me with this section ;) :D

PR #89, opened 2021-09-18
Guess Who by Lena Simma

lenisili:


code/script.js
```diff
@@ -227,99 +229,146 @@ const setSecret = () => {
...
+
 // This function to start (and restart) the game
 const start = () => {
-  // Here we're setting charactersInPlay array to be all the characters to start with
+  winOrLoseWrapper.style.display = "none";
   charactersInPlay = CHARACTERS
-  // What else should happen when we start the game?
+  generateBoard();
+  setSecret();
+  console.log(secret);
```
Fatima-Gamero-Romero, line 239: According to Technigo's teacher ,console.logs should be removed before pushing the projects as it's just for developers to see behind the scenes.

code/script.js
```diff
@@ -227,99 +229,146 @@ const setSecret = () => {
...
+ generateBoard()
 }
 
 // when clicking guess, the player first have to confirm that they want to make a guess.
 const guess = (personToConfirm) => {
-  // store the interaction from the player in a variable.
-  // remember the confirm() ?
-  // If the player wants to guess, invoke the checkMyGuess function.
+  confirm(`Do you really want to guess?`) 
+  if (true){
```
Fatima-Gamero-Romero, line 348: This line is always going to be true regardless of the user choice. So, if the user wants to guess but then, the user for some reason changes his/her mind and press the cancel button, the message should not be shown and instead, the board game. 
So you need only an if statement, no need an else, and only if the user wants to guess, the message will be shown, if not, the board game. 
I am going to copy here my code so you can understand what I mean 😊

```
let confirmation = confirm('Are you really sure you want to guess on this person?'); 
  if(confirmation) {
    checkMyGuess(personToConfirm)
  }  
};
```
// The variable confirmation returns a boolean value, either true if the user accepts or false if the user doesn't want to guess. (It works the same way as the prompt(), whatever the user types, will be the value )
// If the player wants to guess(if confirmation === true), the checkMyGuess function will be invoked.

Fatima-Gamero-Romero (commented):
Well done Lena, your project works well and your code is very well organized and easy to follow! Now it's time to play this game 😉

PR #97, opened 2021-09-18
Julia N guess who

JuliaNiki:
https://guess-whogame.netlify.app 

I started by thinking about the game itself and then planed the different parts and steps of the game. It was challenging at some point to check if the attributes in the questions that user asked matched the attributes of the secret person because some attributes were strings and some were arrays that made it more complicated to compare. But thanks to support from code coaches at Technigo, it went smooth later on after figuring out this step. 

If I had more time I would have added some animation to the cards when flipping them and make a timer that measures how long time each game takes.


camekman (commented):
You have created a great Guess Who game that covers all the requirements! 
Your code is structured in a good way and you are explaining it very well. It makes it easy to read and understand. The styling is very nice and really enjoy the sound effects you added to when the player win or loose. You should be extra proud of adding the timer and the counter for guesses and that you are displaying how many guesses it took the player to win, great job!

PR #101, opened 2021-09-19
Guess who by Madelene Trang week 5

MT-dotse:


nama0027 (commented):
Job well done!
The code is clear and commented well so easy to follow through. 
The game is working quite well and has cool features like winning and loosing sound. great job! 
However, i would like to point out a small bug; 
on start, the game is not responding to click of `find Out` button until and unless player changes the dropdown value before clicking the button. It is because the eventlistener for <select> is on 'change'. One of possible solutions for this  is changing the default selected value to some type of message like 'select from here'
/Naushin

PR #104, opened 2021-09-19
Project 5: Guess Who - Daniel Vestin

dandeloid:


code/script.js
```diff
@@ -188,23 +193,26 @@ const CHARACTERS = [
...
   },
 ]
 
 // Global variables
 let secret
-let currentQuestion
+let currentQuestion 
 let charactersInPlay
 
+const audioWin = new Audio('./assets/applause7.mp3') // win audio
```
A1eksa, line 213: Nice with the sound:)

code/script.js
```diff
@@ -225,26 +233,29 @@ const generateBoard = () => {
...
   secret = charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)]
+  console.log(secret) //----------Logs secret person. CHEAT code!!! :D
 }
 
 // This function to start (and restart) the game
 const start = () => {
-  // Here we're setting charactersInPlay array to be all the characters to start with
-  charactersInPlay = CHARACTERS
-  // What else should happen when we start the game?
+  charactersInPlay = CHARACTERS // Here we're setting charactersInPlay array to be all the characters to start with
```
A1eksa, line 241: I really like how you explain the code with comments, that is really great.

code/script.js
A1eksa, line 213: Maybe unnecessary to put it in a variable? But that is also a solution ofc.

code/style.css
```diff
@@ -92,6 +107,7 @@ select {
 
 .card:hover {
   background-color: var(--secondary);
+  transform: scale(1.05);
 }
 
 .card:hover img,
```
A1eksa, line 113: Like the hover effect, user friendly

A1eksa (commented):
Great job Daniel! I really like your solution. Project meets all the requirements and beyond so great work. I would only say that maybe you have made too many options so you can pinpoint the person trough selection. But that is just maybe unimportant since the point was to practice Java Script.

PR #115, opened 2021-09-19
Guess Who game by Maja L

Maja473:


README.md
```diff
@@ -1,13 +1,11 @@
...
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+This week's project was creating a game of 'Guess Who', where you play against the computer to guess the correct person based on a randomized choice.
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+JavaScript has, in general, been quite difficult to understand so far since it's a lot at once. I've been having trouble getting the 'play again' button to function even though I can't
```
annaester, line 7: I noticed that your button didn't work. You need to hide the winOrLose and show the board again. You can do it in the eventListener for the play again button. In line 370.

code/script.js
```diff
@@ -231,95 +233,138 @@ const setSecret = () => {
 const start = () => {
   // Here we're setting charactersInPlay array to be all the characters to start with
   charactersInPlay = CHARACTERS
-  // What else should happen when we start the game?
+  generateBoard()
+  setSecret()
```
annaester, line 237: Where is the function that sets the secret? Your chosen secret person is not the same as the one thats left when you've filtered everyone else out. You need the to both set the variable and the function so it works with the game. 
```
const setSecret = () => {
}
```

code/script.js
```diff
@@ -231,95 +233,138 @@ const setSecret = () => {
 const start = () => {
   // Here we're setting charactersInPlay array to be all the characters to start with
   charactersInPlay = CHARACTERS
-  // What else should happen when we start the game?
+  generateBoard()
```
annaester, line 236: The selectQuestion function should be up here aswell. You can't press on the find out button if you haven't changed the question. Or did you not want the function to be invoked from the start?

code/script.js
```diff
@@ -231,95 +233,138 @@ const setSecret = () => {
...
 }
 
 // when clicking guess, the player first have to confirm that they want to make a guess.
 const guess = (personToConfirm) => {
-  // store the interaction from the player in a variable.
-  // remember the confirm() ?
-  // If the player wants to guess, invoke the checkMyGuess function.
+  choosePerson = confirm (`Guess on ${personToConfirm}?`)
+  if (choosePerson) {
```
annaester, line 341: Here you have started a conditional but it doesn't lead anywhere? Or what have I missed?

code/script.js
```diff
@@ -231,95 +233,138 @@ const setSecret = () => {
...
 
 // Invokes the start function when website is loaded
 start()
 
-// All the event listeners
+// All event listeners
 restartButton.addEventListener('click', start)
+questions.addEventListener('change', selectQuestion)
+filter.addEventListener('click', checkQuestion)
+playAgain.addEventListener('click', start)
```
annaester, line 370: Here you can add a function that removes the winOrLose - like the display = flex you did above to show it, but now display = none and the board display = flex.

annaester (changes requested):
Hi! Great work! But you need to fix the secret. It doesn't work with the game, look at my comments. The secret should be set after the generateBoard and before start because they don't work together right now. If you consloe.log(secret) in dev tools and then filter out all the other character you will notice that the one you have left to guess on is not the secret.

PR #123, opened 2021-09-21
Guess who week 5

Asivol93:


code/style.css
```diff
@@ -59,13 +66,72 @@ select {
...
+  width: 100%;
+  height: 100%;
+  -webkit-backface-visibility: hidden;
+  backface-visibility: hidden;
+}
+
+.flip-card-back {
+  transform: rotateY(180deg);
+}
+
```
jakobxlindstrom, line 128: Sweet css styling!

code/script.js
```diff
@@ -229,97 +389,139 @@ const setSecret = () => {
...
 const start = () => {
-  // Here we're setting charactersInPlay array to be all the characters to start with
-  charactersInPlay = CHARACTERS
-  // What else should happen when we start the game?
-}
+  charactersInPlay = CHARACTERS;
+  generateBoard();
+  setSecret();
+  numberOfGuesses = 0
+  guessesInTotal.innerText = 0
```
jakobxlindstrom, line 396: Ah, this is what I needed.

code/script.js
```diff
@@ -229,97 +389,139 @@ const setSecret = () => {
...
+  if (suspectCharacter === secret.name) {
+  winOrLoseText.innerHTML = `Victory! You guessed on ${suspectCharacter} which is the right champion!`
+  victorySound.play()
+} else {
+  winOrLoseText.innerHTML = `Defeat! The right champion was ${secret.name}...`
+  defeatSound.play()
+}
+  
+  winOrLose.style.display = 'flex';
+
```
jakobxlindstrom, line 506: nice! the winOrLose div is shown but did you hide the game board?
Bigger question is do we need to hide it if we cant see it anyway? 
Referring to line 318.

code/script.js
```diff
@@ -229,97 +389,139 @@ const setSecret = () => {
...
+  defeatSound.play()
+}
+  
+  winOrLose.style.display = 'flex';
+  
 }
 
 // Invokes the start function when website is loaded
-start()
+start();
```
jakobxlindstrom, line 510: Just realized sometimes you use ; and sometimes no semi-colons. I think it could be because of prettier. 

If I can recommend something it would be to decide what you feel is best for you, to use it or not and that goes for "or '. In prettier you can set what you want to use so that it formats the way you decide! 
Small detail but again makes your code look smooth.

code/script.js
```diff
@@ -229,97 +389,139 @@ const setSecret = () => {
...
+})
+questions.addEventListener('change', selectQuestion)
+findOutButton.addEventListener('click', checkQuestion)
+playAgainButton.addEventListener('click', () => {
+  start()
+  restartSound.play()
+  winOrLose.style.display = 'none';
+}) 
+
+
```
jakobxlindstrom, line 527: Great listeners!

code/script.js
Asivol93: I use prettier and do never add semicolon myself so thank you for giving me the tips to set what my default should be :)

code/script.js
Asivol93: Not sure what you are referring to in line/row 318, it's inte the middle of an object :D do you mean the javascript or is it another row your thinking about?

The code on row 505 displays the win or loose wrapper. If I remove it and try to guess on a character it won't appear..

code/style.css
Asivol93: Thank you! Found a great source on w3 schools: [https://www.w3schools.com/howto/howto_css_flip_card.asp]

code/script.js
jakobxlindstrom: I think I was unclear. But I'm referring to line 505/506 in your code and the row 318 is from the starter code I think.
In the "starter code" they had a comment that said something like "don't forget to show the winorlose screen and also hide the gameboard".

So I was thinking that maybe there should be a `board.style.display ="none"` together with the `winOrLose.style.display = 'flex'`.

So thats what I meant with my second statement about do we really need it because it seems good to just make the winorlose screen appear.

code/script.js
jakobxlindstrom: Great! Then me and you are in the same boat. I dont like the semi colon in javascript. :D

PR #128, opened 2021-09-22
Project-guess-who/AleksandraSafranko

A1eksa:


pdetli (commented):
Good Job 👍  Your site met the requirements! 
It could be more clean and readable (I think you like the console logs as much as I do, therefore you don't want to remove them :)) 
Probably you forgot to add extra features on the html. Obviously it works in that way also but it would be good to see the other features like hair and eyes colors or even more.

PR #139, opened 2022-02-11
Finished project

Dopest0727:


code/style.css
```diff
@@ -1,233 +1,362 @@
...
+     height: inherit;
+     letter-spacing: .1rem;
+     font-weight: 300;
 }
 
-.card:hover {
-  background-color: var(--secondary);
+ .card:hover {
+     background-color: var(--secondary);
+     transition: 1s;
```
Laura-Sjolander, line 190: Mobile mode as well as DevTools responsive mode: I don´t know if it´s something in the hover function or what it is that makes the whole card turn black if you click it (since you are not able to hover on mobile or in responsive mode in DevTools). There is no white text shown "Guess on Nadia" either (as it is on PC-version). I´m sure we will learn how to handle hover functions for mobiles and ipads later on in the course :)

README.md
```diff
@@ -1,13 +1,14 @@
 # Project Name
-
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+Guess who project.
+Pretty selfexplainatory. What i loved about this project was that it made me use a lot of creative tools.
```
Laura-Sjolander, line 3: It´s great to have an informative Readme file, if a tech recruiter happens to stop by and view your code. 
Maybe write that we had one week for the assignment and what kind of starter code we were given.

Laura-Sjolander (commented):
As always, very good looking assignment. You have a really good eye for design and put a lot of effort into the projects not only having a nice functionality, but also looking real nice! 

When I looked at the Netlify page in the DevTools console I get a error message: "Failed to load resource: the server responded with a status of 404 () /favicon.ico:1 "
So I guess there is one icon missing? There is only two comments/questions in the code about the readme and in the css-file for the styling of cards in mobile mode. 

Overall impression: Great function, nice mobile first approach. 
I really liked that you made the game about classmates. Nice feature that it changes background color as well as the sound effects when winning or loosing. 
The added timer and that it counts the questions is fab, always aiming for that black level in every assignment. 🥇

PR #143, opened 2022-02-11
Guess who ? - LEGO edition (Emma Högberg)

emmahogberg88:
https://guess-who-lego.netlify.app/


code/script.js
```diff
@@ -222,104 +306,218 @@ const generateBoard = () => {
...
   // We also need a variable that stores the actual value of the question we've selected.
-  // const value =
-
+  
   currentQuestion = {
-    category: category,
-    // value: value
+    category: category, 
+    value: value,
+    text: text,
```
kolkri, line 337: In the currenQuestion object you have added third property called text, and you use it later to print the text if the secret person has or not has the asked property. Is there a reason you added this third 'text' property and not just used the 'value' when printing out the texts in function filterCharacters?

kolkri (commented):
Really great job and the lego topic is so cool with all the images and styling and a lot of different categories to guess! The game fulfills all the minimum requirements, which were for this project: 
- The board with characters should be generated when the website is loaded.
- A randomly selected person should be set as the secret when the game starts.
- You should be able to select questions to ask about the people and filter the board based on those questions.
- You should also give the player feedback with alerts
- You should be able to guess the secret person and get an answer if it's correct or not.

I was wondering the line 337 in JS file (see the comment below). Also you could have combine some if / else statements, but I understand it is more clear way to make them separately when you have so many different keys&values in the character objects. Overall, really great project and everything works as wanted in the game!

code/script.js
emmahogberg88: I had the value at first, but it was hard for me to find a good way to print it out in the alerts with "good" English, and also being able to use it as an understandable value inside the characters array. That´s why I decided to save the text as well :)

emmahogberg88:
Thank you Kristina! I agree with you. I thought of combining the if/else statements for the alert and the updated charactersInPlay, but when I asked Jennie if I could do so during Town Hall I understand from her that I shouldn´t do that? But I think I might have misunderstood her maybe.. because it isn´t really DRY at the moment and I have seen that other person has combine them and I think it looks good and is clear that way..

PR #147, opened 2022-02-11
project-guess-who

rijad90:


EmmaaBerg (commented):
Really good job, the game works as it should and the blue requirements are fulfilled. The naming in the JS is good and the code was structured. It's good with comments, especially since this is all new and we are still learning. But maybe some comments could be removed to make the code more readable and easy to follow.

PR #150, opened 2022-02-11
guess-who-Frida Axelsson

Nosslexa:


code/script.js
```diff
@@ -225,101 +232,164 @@ const generateBoard = () => {
 // Randomly select a person from the characters array and set as the value of the variable called secret
 const setSecret = () => {
   secret = charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)]
+  console.log('secret person selected')
+  console.log(secret)
```
ariallahyar, line 236: With `console.log(secret)` anyone who goes into Inspector can cheat ;) It's a good tool to have while testing your filtering, but then it should be removed.

code/script.js
```diff
@@ -225,101 +232,164 @@ const generateBoard = () => {
...
+  console.log('question asked', currentQuestion)
 }
 
 // This function should be invoked when you click on 'Find Out' button.
+let count = 0
 const checkQuestion = () => {
-  const { category, value } = currentQuestion
+  count+= 1
+  guessCounter.innerHTML = " "
+  guessCounter.innerHTML += `<p> Guesses made: ${count}</p>`
```
ariallahyar, line 270: I recall you mentioning the counter as a stretch goal you had, nice work! However, the counter should be reset to 0 when the game is restarting.

code/script.js
```diff
@@ -225,101 +232,164 @@ const generateBoard = () => {
...
 const start = () => {
+  console.log('start funktion is called!')
   // Here we're setting charactersInPlay array to be all the characters to start with
   charactersInPlay = CHARACTERS
   // What else should happen when we start the game?
+  // set the secret person -> invoke set Secret funktion
+  setSecret()
+  //show the board on the screen -> invoking generateboard funktion. 
+  generateBoard()
 }
```
ariallahyar, line 249: You could revise` start() `to include the few things from newGame, and remove newGame. It is the same thing, ie drawing up the board of characters, selecting the secret character, and setting the counter to 0. So if you add the two lines plus set count to 0, the event listener on `playAgainButton` can call start. 
```javascript 
winOrLose.style.display = 'none'
board.style.display = 'flex' `
count = 0 // add here to reset your question counter

File: code/script.js [hunk 4]
```diff
@@ -225,101 +232,164 @@ const generateBoard = () => {
...
   // Then invoke filterCharacters
-  if (category === 'hair' || category === 'eyes') {
 
+  if (category === 'hair' || category === 'eyes') {
+    if(value === secret.hair || value === secret.eyes){
+      keep = true
+      filterCharacters(keep)
+    }
+    else{
+      keep = false
```
ariallahyar, line 284: I don't think `keep = false` is being used anywhere because when you call `filterCharacters` you don't take keep as an argument, so it could be deleted here and on line 293

code/script.js
```diff
@@ -225,101 +232,164 @@ const generateBoard = () => {
...
+      filterCharacters()
+    }
   } else if (category === 'accessories' || category === 'other') {
-
+      if (secret.accessories.includes(value) || secret.other.includes(value)){
+      keep = true
+      filterCharacters(keep)
+    }
+    else{
+      keep = false
```
ariallahyar, line 293: Same as previous comment, if keep is unused it doesn't need to be here I don't think.

code/script.js
```diff
@@ -225,101 +232,164 @@ const generateBoard = () => {
...
 
 // Invokes the start function when website is loaded
 start()
 
+//restart game
+const newGame = () =>{
+winOrLose.style.display = 'none'
+board.style.display = 'flex'
+start()
+}
```
ariallahyar, line 388: As mentioned up in the start function, this newGame function is sort of repetitive. You could instead move two lines into the start function and always call start weather you load the page, restart the game or choose to play again after making a guess.

ariallahyar (commented):
Nice work, Frida! Your code is well structured and therefore easy to follow. 

I noticed you included many console.logs with descriptive messages. Printing out messages to the console is a great tool while you're building your projects, but it can later be removed since it does not add any value. 

Good job getting the question counter in there, just remember next time to think about resetting counters if there is the possibility of restarting a game or whatever it might be! :)

code/script.js
Nosslexa: Thanks, completely forgot and will remove.

code/script.js
Nosslexa: thanks, i updated the start function as you suggested and removed the new game function. Also tried to clear the counter but don't get it to work. Any ideas on how to solve this?

ariallahyar: I just tested added this line of code to your start function after you set `count = 0` and it worked!
```javascript
guessCounter.innerHTML = `<p> Guesses made: ${count}</p>`
```
Otherwise, your counter variable is being reset to 0 when you start or restart, but the text being displayed isn't resetting until you submit a question. Hope that helps :)

Nosslexa: yes, it works! Thanks for the help :)

PR #151, opened 2022-02-11
Update is done fulfilling all requirement

eyahya-khan:


js/script.js
```diff
@@ -586,11 +572,20 @@ const guess = (personToConfirm) => {
   // store the interaction from the player in a variable.
     // remember the confirm() ?
     // If the player wants to guess, invoke the checkMyGuess function.
-  const confirmed = confirm(`Do you think the secret person is ${personToConfirm} ?`)
-
-  if (confirmed) {
-    checkMyGuess(personToConfirm)
-  }
+ 
+  Swal.fire({
```
Katarina821, line 576: Nicley done with styling of the confirm-box!

js/script.js
```diff
@@ -520,55 +521,40 @@ const checkQuestion = () => {
...
+   Swal.fire(`Yes! Guess is correct, the person has a ${attribute}.`, `Keep all who has a ${attribute}.`)
 } else {
   sound()
-    alert(
-      `Wrong guess, the person doesn't have a ${attribute}. Remove all that has a ${attribute}.`
-    )
+   Swal.fire(`Wrong guess, the person doesn't have a ${attribute}.`, `Remove all that has a ${attribute}.`)
   } 
 }
   // filter to keep or remove based on the keep variable.
```
Katarina821, line 560: Good job with the filterfunction, interesting to see that it can be written like that

js/script.js
```diff
@@ -439,6 +439,7 @@ let secret, currentQuestion, charactersInPlay
```
Katarina821, line 439: I think if you would write let currentQuestion = { category: 'hair', value: 'brown' } here you would be able to choose that also on the first guess, which you cant as it is now.

Katarina821 (approved):
Hi Eyahya! Nicely done with the project, the site looks great and the code was easy to read. Good that you made comments, that made it easier to follow. I learned a lot by looking at your code! I left you one comment with a suggestion on how to make it possible to choose "brown hair" also on the first guess, right now you cant do that in the game. Otherwise i think it was a really well done project, congrats!

PR #152, opened 2022-02-12
Project 4 - Guess who

toal13:


code/script.js
```diff
@@ -253,73 +258,129 @@ const checkQuestion = () => {
...
 // It'll filter the characters array and redraw the game board.
 const filterCharacters = (keep) => {
   const { category, value } = currentQuestion
   // Show the correct alert message for different categories
   if (category === 'accessories') {
     if (keep) {
-      alert(
-        `Yes, the person wears ${value}! Keep all people that wears ${value}`
-      )
+      Swal.fire(`Yes, the person wears ${value}!`)
```
amandatange, line 286: Here as well would be good to have some comments to explain what this does!

amandatange (commented):
The game works well and the changes to the design are very nice! I like the custom alerts but I think there could be some more commenting of the code in general, and in particular the parts with those alerts. The code is well structured which makes it easier to read, but I really would have liked more comments to  guide me a little!:) You've done a really good job and I definitely think you reach the general requirements, and then some!

PR #155, opened 2022-02-12
Guess Who Week 5, Linnea Frisk

Neaa99:


JenFi72:
Amazing work Linnéa!!! I am very impressed with what you've accomplished here - the games does not only work very well, it looks very good as well. Blue level requirements are met by far, even reaching some of requirements for red and black level as you put your own (very good looking Games of Thrones) theme. Nice that you added sound effect as well. REALLY GOOD JOB!

PR #156, opened 2022-02-12
Project guess who W5 

Mimmisen:
https://guess-who-w5.netlify.app/

### DIFF
--- a/code/index.html
+++ b/code/index.html
@@ -23,6 +23,7 @@
       <h1>Does the person have</h1>
 
       <select id="questions">
+        <option value="" selected disabled>Select attribute</option>
         <optgroup label="hair">
           <option value="brown">brown hair</option>
           <option value="yellow">yellow hair</option>
--- a/code/index.html
+++ b/code/index.html
@@ -3,10 +3,12 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Guess Who</title>
+    <title>Guess Who game?</title>
     <link rel="stylesheet" href="style.css" />
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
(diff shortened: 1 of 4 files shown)


code/script.js
```diff
@@ -218,108 +224,178 @@ const generateBoard = () => {
...
-}
+  add();
+  questionsAsked.innerHTML = `
+  ${counter}`;
+};
 
 // It'll filter the characters array and redraw the game board.
 const filterCharacters = (keep) => {
-  const { category, value } = currentQuestion
+  const { category, value, text } = currentQuestion;
```
tiiliu, line 297: Here I'm not sure what the 'text' inside the const currentQuestion is, I wasn't able to find what it might be referring to?

code/index.html
```diff
@@ -20,28 +22,51 @@
         alt="Guess Who"
       />
 
-      <h1>Does the person have</h1>
+      <h1>Does the person have...</h1>
 
       <select id="questions">
+        <option value="" selected disabled>Select attribute</option>
```
tiiliu, line 28: This is a nice addition and avoids the problem of not being able to choose the default option :)

tiiliu (commented):
Your code was very clean and easy to read. There was good commenting with all the functionalities and I really like the attention to detail, for example with the use of quotes (""). I liked the counterbox feature that you had created and the fact that in your win message you're telling the player how many guesses it took them to win. There wasn't really anything to critique on, the website works well and the styling looks very nice :) Your code fulfills all the blue level requirements and some of the red level ones too. Great job over all! 👏 😊

code/script.js
Mimmisen: I was playing around with using value and text but decided to only use value and removed all 'text' but forgot this one ;)

code/index.html
Mimmisen: Emma helped me with this one and she sad that she got the tip from you so thanks! :D

PR #161, opened 2022-02-12
Guess who project

Katarina821:


code/script.js
```diff
@@ -2,324 +2,393 @@
...
+	img: 'images/Xola.png',
+	fur: 'long',
+	size: 'large',
+	accessories: [],
+	other: []
+}, ]
 // Global variables
+
 let secret
 let currentQuestion
```
vanhaj, line 184: `let currentQuestion = { category: 'fur', value: 'long' }`

vanhaj (approved):
You have done a great job! 
Nice styling and I love that you added dogs as pictures. As mentioned in my previous comment, the pop up alerts are a very cool feature and it adds an extra little introduction to the game. 

One error I noticed was when choosing long fur on the first guess. You can easily fix that by adding a code to the `let currentQuestion` on line 184.  (see my comment there how it could be written)

Awesome work, couldn't have done it better!

PR #165, opened 2022-02-13
Let's Play Guess who!

Kras053:


code/script.js
```diff
@@ -222,104 +226,169 @@ const generateBoard = () => {
...
 }
 
-// If you confirm, this function is invoked
-const checkMyGuess = (personToCheck) => {
-  // 1. Check if the personToCheck is the same as the secret person's name
-  // 2. Set a Message to show in the win or lose section accordingly
-  // 3. Show the win or lose section
-  // 4. Hide the game board
+//here we show the WinOrLose section and hide the game board
+winOrLose.style.display = "block"
```
SteppbySteph, line 373: In the CSS-file this section is set to 'flex'. You can change the current placement by typing 'flex' instead of 'block' to see how it looks like!

code/script.js
```diff
@@ -222,104 +226,169 @@ const generateBoard = () => {
...
 // setting the currentQuestion object when you select something in the dropdown
 const selectQuestion = () => {
+  // This variable stores what option group / category the question belongs to
   const category = questions.options[questions.selectedIndex].parentNode.label
 
-  // This variable stores what option group (category) the question belongs to.
-  // We also need a variable that stores the actual value of the question we've selected.
-  // const value =
+  // This variable stores the actual value of the question you selected within the category
+  const value = questions.options[questions.selectedIndex].value
```
SteppbySteph, line 253: Good one! You could also write this with less code as value = questions.value.

code/script.js
```diff
@@ -222,104 +226,169 @@ const generateBoard = () => {
...
-// This function should be invoked when you click on 'Find Out' button.
+
+// This function is invoked when you click on 'Find Out' button
 const checkQuestion = () => {
   const { category, value } = currentQuestion
 
-  // Compare the currentQuestion details with the secret person details in a different manner based on category (hair/eyes or accessories/others).
-  // See if we should keep or remove people based on that
-  // Then invoke filterCharacters
+  // Compares the currentQuestion details with the secret person details based on category hair/eyes (string) or accessories/others (array)
```
SteppbySteph, line 267: Very good explanation to why the categories are divided into two groups.

code/script.js
```diff
@@ -222,104 +226,169 @@ const generateBoard = () => {
...
-  // 2. Set a Message to show in the win or lose section accordingly
-  // 3. Show the win or lose section
-  // 4. Hide the game board
+//here we show the WinOrLose section and hide the game board
+winOrLose.style.display = "block"
+board.style.display = "none"
 }
 
+// this section is for clicking the restart button, not working unfortunately
+playAgainButton.addEventListener = ('click', (start) => {
```
SteppbySteph, line 378: It seems you have tried to make a function out of an addEventlistener. 
Try playAgainButton.addEventListener('click', start) where "start" is the function that is invoked when you click on the playAgainButton. Try moving your code (row 379 - 380) to your start function instead. Hopefulle this will solve the problem.

PR #179, opened 2022-02-13
kolkri

kolkri:


qwer-ctrl (commented):
Super nice game! I think your js was very easy to follow and I don't have any questions of that. On this project, your CSS styling is THE thing. Fun idea and it was hard to play, because at some point I can't quess on the basis of their looks anymore :D Sound effects, correct value inside if the guess is wrong, counter and timer was little things that set you apart from others. I say this only if you didn't notice (because first I have same problem on my project) but when I find out a question and then scroll down, at right side on a page is a empty white box :)

But very good! One of my favourite guess who games in our class 👍

PR #184, opened 2022-02-13
Blue level commit

JensBergqvist:


code/script.js
```diff
@@ -222,104 +223,141 @@ const generateBoard = () => {
...
+} 
+    
+  
+
 
 // It'll filter the characters array and redraw the game board.
 const filterCharacters = (keep) => {
   const { category, value } = currentQuestion
-  // Show the correct alert message for different categories
+  // Alerts
```
HWallberg, line 286: Great that you used headlines like this, makes it easy to follow

HWallberg:
You only uploaded one file; the js-file. But no html or css.

HWallberg (commented):
Very clear code. Well done. It would have been fun to see you change up the styling a bit. And also add some more options in the html maybe?

PR #187, opened 2022-02-13
Week 5- Guess who- Laura

Laura-Sjolander:
Here is my hand in for week 5. Guess who game. /Laura Sjölander

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -1,5 +1,6 @@
 // All the DOM selectors stored as short variables
 const board = document.getElementById('board')
+//const card = document.getElementById('card') måste ha korten i klassen card? 
 const questions = document.getElementById('questions')
 const restartButton = document.getElementById('restart')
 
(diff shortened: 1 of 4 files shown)


code/script.js
```diff
@@ -222,104 +225,130 @@ const generateBoard = () => {
...
+// If restarting the win or loose screen will be hidden
 const start = () => {
-  // Here we're setting charactersInPlay array to be all the characters to start with
   charactersInPlay = CHARACTERS
-  // What else should happen when we start the game?
+  generateBoard()
+  setSecret()
+  selectQuestion()
+  winOrLose.classList.remove("show-win-or-lose");
 }
```
jessand77, line 241: Here you can add `board.style.display = 'flex'` in order to display the board again if you hide it in the checkMyGuess function.

jessand77 (commented):
Very good work with this project! The code is easy to follow, the comments are very good and so are the names of the variables, parameters, classes et cetera. The game works as it should and the project follows the general and the blue requirements (as well as some of the red requirements). One thing that you can change if you want is to hide the game board completely when the win or lose section is displayed. See my comments on the start function and the checkMyGuess function - hopefully it will work with these changes :-)

PR #189, opened 2022-02-13
Guess Who- Sara Boström Bundock

bostrombundock:


code/script.js
```diff
@@ -222,104 +223,121 @@ const generateBoard = () => {
...
 
-// If you confirm, this function is invoked
+  if (thePlayerGuess) {
+    checkMyGuess(personToConfirm) 
+ 
+  } else {
+    alert (`Sorry, try again!`)
+
+  }  
+}
```
sieurin, line 323: It seems like you use different spaces and indentations in your project. You can use prettier code formatter that helps clean up your code, so you don't have for example unnecessary spaces or improper indentations :)

sieurin (commented):
Great work Sara!  Your site works as it should, no bugs and you have met the blue requirements. Well done! You should be proud! 

The only note would  be that your code could be tided up a bit (there was some random spaces and improper indentations), and small comments - merely suggestions. 

I would also suggest to comment your scripts a bit, that helps to follow your code and understand what the different functions do (especially when the projects become bigger).

I don’t see any big changes to do, so keeping this review pretty short :)

PR #191, opened 2022-02-13
Week 5 - project Guess Who

mathildakarlsson:
This week I focused on the most important things. Even though I wanted to change the characters and so on, I feel happy that I completed the task and learned a lot about javascript! I tried to do as much commenting as I could, including keeping the helpful default comments, to understand if I'll go back and watch this project in a later state. 

The community on slack and stack overflow has been so helpful this week and I realized other students struggled with the same things as I did, which made it easier to find different inputs on solutions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
-# Project Name
-
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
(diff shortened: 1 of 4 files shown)


mistscale (commented):
I'm very impressed with your code and how you managed to solve it. The game works well and I couldn't really find any improvements since the code was clear and easy to follow. Good job!

PR #193, opened 2022-02-20
Guess who game - Lovisa

LovisaBrorson:


code/script.js
```diff
@@ -226,7 +229,7 @@ const generateBoard = () => {
 // Randomly select a person from the characters array and set as the value of the variable called secret
 const setSecret = () => {
   secret = charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)]
-  
+  console.log(secret)
```
sukiphan97: Remember to remove the console.log, otherwise, player can open the inspector and find out the secret person 😉

code/script.js
```diff
@@ -235,20 +238,18 @@ const start = () => {
...
   // Here we're setting charactersInPlay array to be all the characters to start with
   // What else should happen when we start the game?
 }
 
 // setting the currentQuestion object when you select something in the dropdown
 const selectQuestion = () => {
   const category = questions.options[questions.selectedIndex].parentNode.label
-  const value = questions.value
- 
+  const value = questions.options[questions.selectedIndex].value
```
sukiphan97, line 248: Another way to get the value is 
`  const value = questions.value;`

Because the all the `options `are values of `select`, in this case, it's unnecessary to attach the  `options[questions.selectedIndex]`

code/script.js
```diff
@@ -257,47 +258,72 @@ const selectQuestion = () => {
...
+  const {category, value} = currentQuestion
 
   // Compare the currentQuestion details with the secret person details in a different manner based on category (hair/eyes or accessories/others).
   // See if we should keep or remove people based on that
   // Then invoke filterCharacters
-  let keep = false
 
   if (category === 'hair' || category === 'eyes') {
-    keep = value === secret[category];
+    if (value === secret.hair || value === secret.eyes) {
```
sukiphan97, line 268: Because you already defined category === 'hair' || category === 'eyes'
in the line 268. A shorter way to write this is
if (value === secret.category )

code/script.js
```diff
@@ -257,47 +258,72 @@ const selectQuestion = () => {
...
+      filterCharacters(true)
+    } else {
+      filterCharacters(false)
+    }
+
   } else if (category === 'accessories' || category === 'other') {
-    keep = secret[category].inCludes(value);
+    if ((secret.accessories).includes(value)) {
+      filterCharacters(true)
+    } else if((secret.other).includes(value)) {
```
sukiphan97: As I mentioned the code above (line 269), we can combine all this code in one line. Instead of listing out all the catagories such as accessories and other. 
`if ((secret.accessories).includes(value))`
and `else if((secret.other).includes(value))`

We can do like this: 
`   if (secret[category].includes(value)) {
      filterCharacters(true);
    } else {
      filterCharacters(false)
    }`
    
    Now the category is either accessories or other

sukiphan97 (commented):
Hi Lovisa,

Overally, the codes look great and work perfectly. There are some minor changes I suggested there, you can have a look and let me know if there is something to clarify. I don't have any review about the html and styling, as they are perfect. 

Just small example for sound if you want to try: 
`const magicSound = new Audio ("audio/magic-sound.wav");
 magicSound.play()
`

sukiphan97:
Review

PR #196, opened 2022-09-01
Fredrik – Project Guess Who?

karlfredrikholm:
vem-dar.netlify.com


Tara78:
Hej Fredrik, 

Your code is excellent, It is nothing I can add. Just to say that well done !!!

PR #197, opened 2022-09-02
Project Guess who

agadman:
https://remarkable-sundae-d1bafe.netlify.app/

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -2,6 +2,7 @@
 const board = document.getElementById('board')
 const questions = document.getElementById('questions')
 const restartButton = document.getElementById('restart')
+const filterButton = document.getElementById('filter') // ANNIKA
 
 // Array with all the characters, as objects
 const CHARACTERS = [
@@ -224,8 +225,9 @@ const generateBoard = () => {
 
 // Randomly select a person from the characters array and set as the value of the variable called secret
 const setSecret = () => {
-  secret = charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)]
-}
+  secret = charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)];
+  console.log(secret);
+};
 
 // This function to start (and restart) the game
 const start = () => {
@@ -234,6 +236,7 @@ const start = () => {
   // What else should happen when we start the game?
   generateBoard(); //ANNIKA 1. calling the function show the board
   setSecret();    // ANNIKA 2. setting a secret person at start
+  selectQuestion(); //ANNIKA
 }
 
 // setting the currentQuestion object when you select something in the dropdown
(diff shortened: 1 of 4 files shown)


code/script.js
```diff
@@ -256,39 +268,78 @@ const checkQuestion = () => {
...
+    } else {
+      alert(
+        `No, the person doesn't have ${value} eyes! Remove all people with ${value} eyes`
+      )
+      charactersInPlay = charactersInPlay.filter((person) => person[category] !== value);
+    }
+  } else if (category === 'accessories') {
+    if (keep) {
+      alert(
+        `Yes, the person has ${value}! Keep all people that has ${value}`
```
JosefineEngdahl, line 317: One thing I also struggle with is to get the grammatics to work depending on if you choose glasses or hat, ”The person wears glasses” sounds correct but ”the person wears hat” sounds a bit strange (shouldn’t it be ”the person wears a hat”?). Haven’t solved this one yet, so let me know if you do :).

code/script.js
```diff
@@ -224,29 +228,37 @@ const generateBoard = () => {
...
 // This function to start (and restart) the game
 const start = () => {
   // Here we're setting charactersInPlay array to be all the characters to start with
   charactersInPlay = CHARACTERS
   // What else should happen when we start the game?
+  generateBoard();  // invokes the function show the board
+  setSecret();      // invokes the function that sets a secret person
+  selectQuestion(); // invokes the function to be able to select questions
+
+  board.style.display = "flex"      // adding flexbox to this id
```
JosefineEngdahl, line 243: Nicely solved with the "board=flex" and "winOrlose=none "in this section. I had to create a new const named "Playagain" to make it work, this is a better solution!

code/script.js
```diff
@@ -224,29 +228,37 @@ const generateBoard = () => {
...
+};
 
 // This function to start (and restart) the game
 const start = () => {
   // Here we're setting charactersInPlay array to be all the characters to start with
   charactersInPlay = CHARACTERS
   // What else should happen when we start the game?
+  generateBoard();  // invokes the function show the board
+  setSecret();      // invokes the function that sets a secret person
+  selectQuestion(); // invokes the function to be able to select questions
```
JosefineEngdahl, line 241: I'm missing the "selectQuestion-function in my code, but it still works. I THINK it's because I added this in the HTML:        <option value="" disabled selected>Select your guess</option> before "Label hair". But you manage to make it work without updating the HTML, nice!

code/script.js
```diff
@@ -256,39 +268,78 @@ const checkQuestion = () => {
...
+      charactersInPlay = charactersInPlay.filter((person) => person[category] !== value);
     }
-  } else if (category === 'other') {
+  } else if (category === 'eyes') {
     // Similar to the one above
+    if (keep) {
+      alert(
+        `Yes, the person has ${value} eyes! Keep all people that has ${value} eyes`
+      )
+      charactersInPlay = charactersInPlay.filter((person) => person[category] === value);
```
JosefineEngdahl, line 307: I have my "charactersInPlay=charactersInPlay.filter((person) =>....... " in a separate section, you said that didn't work for you. The difference between our code is that I decided that "keep = true", maybe that's why it works for me? But I still think it looks nice to do it your way!

JosefineEngdahl (approved):
All requirements are met and also you did some very nice CSS-styling, great job! Also very well comment, it’s easy to follow your code. You ROCK! 🎖🌟

PR #200, opened 2022-09-02
w2-ProjectGuessWho-SofieRydmark

SofieRydmark:
Link to project: https://project-guesswho-sofierydmark.netlify.app/


code/script.js
```diff
@@ -215,111 +222,213 @@ const generateBoard = () => {
...
+// This function is to start (and restart) the game
 const start = () => {
-  // Here we're setting charactersInPlay array to be all the characters to start with
-  charactersInPlay = CHARACTERS
-  // What else should happen when we start the game?
+  charactersInPlay = CHARACTERS;
+  generateBoard();
+  setSecretCharacter();
+  selectQuestion();
+  alert('lets play!')
```
KajaWilbik: Is that alert making display a place to put a name?

KajaWilbik (approved):
Your Guess Who game looks very good. The sound when you guess a bad character is amazing. There is only one small detail that I realized does not work properly on the desktop. When you guess a person and move to a win-or-lose screen you can see a white area and a half of the "Find out" button. Was that a plan in the design?  On the phone looks like everything is ok. 

Great job!

PR #216, opened 2022-09-04
guess-who- project-Johanna-Rosenholm

RosenholmJohanna:
https://aquamarine-caramel-104052.netlify.app/


code/script.js
```diff
@@ -222,104 +243,151 @@ const generateBoard = () => {
...
+      }
 
-  } else if (category === 'accessories' || category === 'other') {
-
+  } else if (category === 'accessories' || category === 'other') 
+    { if (secret[category].includes(value)) {
+    keep = true
+  } else {
+    keep = false
   }
```
Iracuzzi, line 302: Nice form to make work the "or" option, with the "double if" is a great way to make it work! Good job!

code/script.js
```diff
@@ -222,104 +243,151 @@ const generateBoard = () => {
...
+  }
 }
+start()  //HTML
+
+// Event listeners
+restartButton.addEventListener('click', start)  // restart current game
+questions.addEventListener('change', selectQuestion)  // dropdown with selections 
+findOutButton.addEventListener('click', checkQuestion && updateGuesses) 
+playAgainButton.addEventListener('click', start) // back to homepage)
+playButton.addEventListener('click',closeOverlay);
```
Iracuzzi, line 391: I think here where better to put a 'submit' listener, because if you press the key enter in the name input bar, the browser refresh. And put the input bar id better than the button.

Iracuzzi (commented):
This is a great job you did here! You made so much easy to understand the code and don't need a big lines of code, really a good job here!

PR #217, opened 2022-09-04
Week 2, Guess Who, Sofia A

SofiaErikaIngeborg:
https://shiny-scone-950594.netlify.app/


Archkrull (commented):
The code is easy to understand and it meets all of the requirements, and functions as it should!
A thing we could do to improve is to delete a lot of comments and add more descrition to the functions. It is a bit messy with so many comments and code snippets commented out but still kept.

Overall, easy to understand, does its job but could be cleaned up a bit more. Good job!

PR #220, opened 2022-09-04
Project-guess-who-David-Ballester

Fonnt:


ElinSegelov (commented):
Great job David! I really like the sweet alerts! They make it look so much nicer!

Playing through the game I did com across a bug. I had asked questions to come down to only Jean and I asked if the person was a smoker. The warning for only one question left showed up and the Jean was filtered away leaving the board empty.

Other than that, maybe remove the console logs when you see that everything is working properly.
Nice work!

PR #223, opened 2022-09-04
Project Guess who

jesshansson:
Technigo week 2 project


code/index.html
```diff
@@ -16,27 +17,44 @@
 
       <img
         class="guess-who-icon"
-        src="images/guess-who-bubble.png"
+        src="images/guess-who-bubble4.png"
```
jessikalind, line 20: This logo looks great with the styling (colors etc) that you have chosen!

code/script.js
```diff
@@ -224,102 +228,135 @@ const generateBoard = () => {
...
+
+
+// ALL THE EVENTLISTENERS
+// When the click event is triggered, the start function is invoked
+restartButton.addEventListener('click', start);
+
+//When the dropdown is changed, the selectQuestion-function is invoked. 
+questions.addEventListener('change', selectQuestion);
+
+filter.addEventListener('click', checkQuestion);
```
jessikalind, line 353: This is fairly self-explanatory, but since you added such nice comments explaining the other event listeners, perhaps a comment could have been added here as well.

jessikalind (commented):
Great work on week 2, Jessica! I'm really impressed with the code - everything seems to work perfectly! And it is great that you managed to solve the problem with the restart-button after the hand-in. Like you will see in my last comment, it got me thinking about your playAgain-solution. Perhaps, now that you have moved the generateBoard() and setSecret()-functions to the start()-function, you could have invoked the start()-function on the playAgain-button as well. Everything works fine as it is, so there is no need. It would however have shortened the code slightly (being a novice of great proportions, I am however not a 100% that this would work 😇)

I also love that you have changed the icon that shows in the tab in the browser to the ?-icon. Very nice touch! The styling also looks great (love the color-scheme!).

PR #228, opened 2022-09-04
Joel Öhman Guess Who project week 2

joeohm:
https://joels-guess-who.netlify.app/

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -1,17 +1,29 @@
 // All the DOM selectors stored as short variables
-const board = document.getElementById('board')
-const questions = document.getElementById('questions')
-const restartButton = document.getElementById('restart')
+const board = document.getElementById('board');
+const questions = document.getElementById('questions');
+const restartButton = document.getElementById('restart');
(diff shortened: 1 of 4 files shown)


code/script.js
```diff
@@ -218,108 +226,205 @@ const generateBoard = () => {
...
+    if (keep) {
+      charactersInPlay = charactersInPlay.filter(person =>
+        person[category].includes(value)
+      );
+    } else {
+      charactersInPlay = charactersInPlay.filter(
+        person => !person[category].includes(value)
+      );
+    }
+  }
```
lindanorberg, line 386: Could maybe the filtering be written in the if-statement above where you alert the messages? To shorten the code a little bit.

lindanorberg (commented):
Nice work! I am able to understand the code, it would maybe be helpful with some more comments describing the functions but most of the code is self-explanaroty. 

Everything I tested worked just as I think it should be and the code follows the general and blue requirements, and the red and black ones as well.

PR #230, opened 2022-09-04
Guess who by Jessika Lind

jessikalind:


code/script.js
```diff
@@ -222,104 +227,189 @@ const generateBoard = () => {
...
+  console.log('secret person;', secret) //This console.log allows me to see who the secret person is for each game. 
 }
 
-// This function to start (and restart) the game
+// Function: starts the game. 
 const start = () => {
-  // Here we're setting charactersInPlay array to be all the characters to start with
-  charactersInPlay = CHARACTERS
-  // What else should happen when we start the game?
+  charactersInPlay = CHARACTERS // Here we're setting charactersInPlay array to be all the characters to start with
```
brucette, line 238: The start function could be extended by including the `generateBoard()` and `setSecret()` functions inside it. This way all you would need to run your program is call `start()` on row 391. 
It may also be the reason the event listeners on `playAgainBtn` and `restartButton` don't work (if you tried to call start upon click) - since the only thing that happens now with `start()` is that the character array is assigned to charactersInPlay.

code/script.js
```diff
@@ -222,104 +227,189 @@ const generateBoard = () => {
...
 }
 
-// setting the currentQuestion object when you select something in the dropdown
+// Function: selectQuestion sets the currentQuestion when the user selects something in the dropdown (ie when the user formulates his/her question).
 const selectQuestion = () => {
-  const category = questions.options[questions.selectedIndex].parentNode.label
+  console.log('This is the question that I want answered')
+  const category = questions.options[questions.selectedIndex].parentNode.label // This variable stores what option group (category) the question belongs to.
+  const value = questions.value // This variable stores the actual value of the quesiton we've selected in the dropdown.
```
brucette, line 245: Great, I just learnt that you can write this so shortly, I wrote out the whole `const value = questions.options[questions.selectedIndex].value`
So thanks for that :)

code/script.js
```diff
@@ -222,104 +227,189 @@ const generateBoard = () => {
...
-  currentQuestion = {
-    category: category,
-    // value: value
+    else if (category === 'other') {
+      console.log('other:', value)
+      currentQuestion = {
+        category: category,
+        value: value 
+      }
+     }
```
brucette, line 277: Here you could keep the original code, where the category and value for that current question is stored in the global variable (row 209) that is updated whenever the user asks a question:
`currentQuestion = {
    category: category,
    value: value
    }`
    This way all the if else statements could be deleted.

code/script.js
```diff
@@ -222,104 +227,189 @@ const generateBoard = () => {
...
-  // Compare the currentQuestion details with the secret person details in a different manner based on category (hair/eyes or accessories/others).
-  // See if we should keep or remove people based on that
-  // Then invoke filterCharacters
   if (category === 'hair' || category === 'eyes') {
-
-  } else if (category === 'accessories' || category === 'other') {
-
+    console.log(category, value) // Daniels förslag 
+    if (value === secret.hair || value === secret.eyes) {
+      keep = true
```
brucette, line 290: Since an object's property's value can be accessed with square brackets, this could be shortened as follows:
`if (value === secret[category])` to avoid having to compare the hair and eyes separately (category will be storing either 'hair' or 'eyes', depending on what the player has selected). 
Same goes for row 303. 

I also don't think you need to declare `keep` at all - either as a global variable above or in this function - since it is just a placeholder for `filterCharacters()`. 
I think once you call `filterCharacters()` the computer will check if there is or isn't anything inside the parenthesis and then follows the code accordingly..(this is my unprofessional explanation without correct terminology, but it works in my code that's why I know :D )

code/script.js
```diff
@@ -222,104 +227,189 @@ const generateBoard = () => {
...
+    else {
+      keep = false
+      filterCharacters() //samma som ovan 
+      console.log('You are not on the right track!')
+    }
   }
 }
 
-// It'll filter the characters array and redraw the game board.
+// Function: filterCharachters filters the charachters in the array and redraws the board. The redraw will depend on the user's question.
```
brucette, line 316: Similarly to `selectQuestion()`, `filterCharacters()` could be shortened a bit by having the conditionals for eyes and hair together by using the `category` variable as well, e.g. : 
if (category === 'accessories') {
    if (keep) {
      //keep as you have written the code 
    } 
    else {
       //keep as you have written the code
    }
  } 
  else if (category === 'other') {
    if (keep) {
       //keep as you have written the code
    } 
    else {
       //keep as you have written the code
    }
  } 
  else {
    if (keep) {
      alert(`Yes, the person has ${value} ${category}! Keep all people with ${value} ${category}`);
      charactersInPlay = charactersInPlay.filter((person) => person[category] === value);
    } 
    else {
      alert(`No, the person doesn't have ${value} ${category}! Remove all people with ${value} ${category}`);
      charactersInPlay = charactersInPlay.filter((person) => person[category] !== value);
    }

code/script.js
```diff
@@ -222,104 +227,189 @@ const generateBoard = () => {
...
+      alert(`No, the person doesn't have a ${value}! Remove all people that have a ${value}`)
+      charactersInPlay = charactersInPlay.filter((person) => !person[category].includes(value))
+    }
+  } 
+  console.log(charactersInPlay)
+  generateBoard()
 }
 
-// when clicking guess, the player first have to confirm that they want to make a guess.
+// Function: the user confirms his/her guess and the function checkMyGuess is invoked.
```
brucette, line 370: This function is missing the control for when a user cancels their guess, it still proceeds to the revealing the secret person.

brucette (commented):
Nice touch with the instructions for the game included as well as the link to some nostalgic pictures of the original game with further instructions. The game seemed to work flawlessly until I found a bug at the final minute ;) - when you click the guess button on a person, and then click cancel instead of confirming the guess, the game still proceeds to the end and reveals the secret person. 

As previously, you have good comments explaining each piece of code to the reader and making it easy to follow. As you also mentioned, some of the console.logs could be removed to make it more clean once you have tested and debugged, also for the tech savvy user not to find out the answer! Some variables at the top are maybe so obviously named that they don't need further comments to clarify.
Some of the if-else statements can be shortened to avoid repetition, but all in all very good code for just being week2 of your javascript journey. 
I hope my comments make sense, just let me know if you wanna discuss further. 

Well done!
//Tina

code/script.js
jessikalind: Great, thank you!

code/script.js
jessikalind: I totally forgot about making something happen when the user presses "cancel". Thank you for informning me of this! ☺️

PR #246, opened 2022-09-04
end project guess who

fannycardell:


code/script.js
```diff
@@ -222,86 +223,148 @@ const generateBoard = () => {
...
+      alert(`No, the person doesn't ${value}! Remove all people that is ${value}`)
+      charactersInPlay = charactersInPlay.filter((person) => !person[category].includes(value))
+    }
+    console.log('test five filteroutwithalerts')
+  }
+}
 }
+}
+
+generateBoard()
```
malmen237, line 367: This have ended up at the wrong side of the curly bracket, it should be before the one at line-365. Probably one of the things that made the board stop working, you now have two commands that say the same thing but at different places. When they clash, nothing works.

code/script.js
```diff
@@ -222,86 +223,148 @@ const generateBoard = () => {
...
+    else if (category === 'other') {
+        console.log('other', value)
+        currentQuestion = {
+          category: category,
+          value: value
+        }  
+      }
+    }
+    
+selectQuestion()
```
malmen237, line 280: This have ended up at the wrong side of the curly bracket, it should be before the one at line-278.

malmen237, line 280: I think you might have hade a slight head-spin about this and made it more complicated than it needs to be. At the moment the browser doesn't really know what to do and therefor does nothing. 
This is how I solved it in my code:
````
   // The function that sets the currentQuestion object when you select something in the dropdown
        const category = questions.options[questions.selectedIndex].parentNode.label;
        const value = questions.options[questions.selectedIndex].value;

        currentQuestion = {
            category: category,
            value: value
        }
        checkQuestion();    // Invokes the checkQuestion-function 
        }
````

malmen237 (changes requested):
I know you've had a really har time with this project and I hope that some of my comments can help you solve your issues.

A tip is also to try and keep the indentation, then it's easier to spot it when something goes wrong and gets misplaced. It also makes the code less "messy" and it will probably make it easier for you to read.

Just keep a cool head and think: It's just code and it's all fixable, I just have to figure out how!

PR #254, opened 2023-02-17
Finished blue level

cpz-crafts:
https://verdant-ganache-055aca.netlify.app/

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -2,6 +2,10 @@
 const board = document.getElementById('board')
 const questions = document.getElementById('questions')
 const restartButton = document.getElementById('restart')
+const findOutButton = document.getElementById('filter')
+const playAgainButton = document.getElementById('playAgain')
+
+
 
 // Array with all the characters, as objects
 const CHARACTERS = [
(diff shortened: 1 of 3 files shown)


code/script.js
```diff
@@ -225,101 +173,213 @@ const generateBoard = () => {
...
+time--;
+}, 1000);
+
+
+  // background song option
+
+ const muteSound = () => {
+   if (backGroundSound.muted === true) {
+     backGroundSound.muted = false
+     backGroundSound.play()
```
JosefinRobertsson, line 367: Perhaps a backGroundSound.loop = true to make the music keep going unless the player mutes it again?

JosefinRobertsson (commented):
Nice feature that you can choose whether you want music to play or not. I hit Restart a few times instead of “Find out”, so for less attentive players like myself maybe add an alert to Restart or move it away from the Find Out-area?  Would it be possible to make use of the space on the sides to show the whole board on desktop even when it’s full, and get rid of the need to scroll? 
Lovely, coherent styling and theme! Very good looking game, and the studio Ghibli theme is nicely paired with the retro arcade style fonts and contrasts. Beautiful color choices, and a nice touch with the gif at the end. The artwork is also beautiful! Maybe a footer where you can credit both the artist and yourself? You made it a bit more your own by adding the cap on number of questions too. Easily navigated script with aptly named variables. Great job!

PR #255, opened 2023-02-17
Pull request Guess Who by LT

khleoth:


code/script.js
```diff
@@ -2,59 +2,51 @@
 const board = document.getElementById('board')
 const questions = document.getElementById('questions')
 const restartButton = document.getElementById('restart')
-
```
EvaSjoholm, line 5: Minor detail but, keeping your functions all together just for a easier read perhaps.

code/script.js
```diff
@@ -204,6 +212,17 @@ const CHARACTERS = [
 let secret
 let currentQuestion
 let charactersInPlay
+let userName
+const winTrack = new Audio('./images/woohoo.mp3')
```
EvaSjoholm, line 216: Fun input with the audio!

EvaSjoholm (commented):
Well done with the stretch-goals, all functions: name input for the user, the guess counter as well as timer where accurate. Good audio choice during win! 
Only feedback I could give is keeping the JS a bit more "cleaned up", using one row for same code/function ones the page is working as pleased. 
Good job Leo!

code/script.js
khleoth: I don't get it, they are all gathered at top with the rest of the getElemenById?

I think what you are referring to might be in some other section of the code?

PR #257, opened 2023-02-18
Jennifer's guess who project

jenster5:
https://guess-who-game-technigo.netlify.app/


NinaWald (commented):
Good work! Tried to review your code comparing with my code, not that good at reviewing yet so take this with a pinch of salt as we say in sweden!

PR #264, opened 2023-02-18
Guess who - Ajmal

AjmalZ:


SaraBat (approved):
* The code is well organized and commented, easy to read and follow the thought process of the person who wrote it. 
* Code is well grouped, it doesn’t look like there’s much opportunity to shorten it. 
* All works. A random person on the board is chosen at each turn. When clicking attributes the characters are filtered down. When taking a guess the player can either win or loose. The option is given to restart the game. 
* 100% follows project guidelines

PR #268, opened 2023-02-19
Joanna Philips : Guess Who

Joanna-Philips:
View the deployed site here : https://guess-who-joannaphilips.netlify.app/


code/script.js
```diff
@@ -231,95 +260,201 @@ const setSecret = () => {
...
+    questionsAskedDisplay.innerText = countQuestionsAsked
+
+    if (category === 'hair') {
+      if (keep) {
+        alertify.alert(gameName,
+          `Yes, the suspect has ${value} hair! Keep all suspects that have ${value} hair.`
+        )
+      } else {
+        alertify.alert(gameName,
+          `No, the suspect doesn't have ${value} hair! Remove all suspects that have ${value} hair.`
```
noraterbocs, line 327: You could merge the hair and eyes conditions by  writing first:
`if (category === 'hair' || category === 'eyes')` and then changing the alert message like this:
``No, the suspect doesn't have ${value} hair! Remove all suspects that have ${value} ${category}.``

noraterbocs (commented):
I really like that you made the design and the theme different from the example we got for this project. I think you have a good eye for the design. 
Regarding the JS code I think you solved it well. Some minor points that I would mention is simplifying the filterCharacters function and maybe disabling the acuse button for those cards that have already been filtered out. 
Otherwise, very good job! 👍

PR #273, opened 2023-02-19
project-guess-who-MaryZaia

MaryZaia:
Link to the site: https://heroic-cassata-727e0b.netlify.app

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -2,6 +2,11 @@
 const board = document.getElementById('board')
 const questions = document.getElementById('questions')
 const restartButton = document.getElementById('restart')
+const findOutButton = document.getElementById('filter')
+const winOrLose = document.getElementById('winOrLose')  
+const winOrLoseText= document.getElementById('winOrLoseText')
+const playAgainButton = document.getElementById('playAgain')
+
 
 // Array with all the characters, as objects
 const CHARACTERS = [
@@ -201,9 +206,12 @@ const CHARACTERS = [
 ]
 
 // Global variables
-let secret
+
 let currentQuestion
 let charactersInPlay
+let secret
+
+
 
 // Draw the game board
 const generateBoard = () => {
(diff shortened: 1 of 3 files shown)


code/script.js
```diff
@@ -316,10 +343,24 @@ const checkMyGuess = (personToCheck) => {
...
+
+  let messgage =""
+  if (secret.name ===personToCheck){
+    alert("Holy Moly you guessed correctly!")
+  } else {
+    alert ("Woopsy daisy better luck next time")
+  }
+  winOrLose.innerHTML=message
+  boardWrapper.classList.add("board-wrapper-inactive")
+  winOrLose.classList.add("win-or-lose-wrapper-active")
```
khleoth, line 355: One suggestion to another alternative, making the game being forced to restart upon completion(win/loss) is that you make the right html elements be hidden/shown and likeswise when afterwards, calling for a restart.
instead of using prompt, the screen could be forced to swap to a restart button only with win/loss-text above.

 winOrLose.style.display = "flex"
    board.style.display = "none"
    winOrLoseText.style.display = "block"
is one easy way to implement css commands into JS for example which can be written as is within a function.

code/script.js
```diff
@@ -316,10 +343,24 @@ const checkMyGuess = (personToCheck) => {
...
 }
 
 // Invokes the start function when website is loaded
-start()
+start(selectQuestion())
```
khleoth, line 359: Is it necessary to call upon selectQuestion within the start function? It is allready in the start function to begin with?

khleoth (commented):
Well done Mary!
Another week, another project.
It seems like you are just a little short on the last iteration for the project to meet the requirements as I interpret it.
Iteration 6: 
- Show the win or lose section
- Hide the game board

I included a partial suggestion, feel free to reach out if I've had a hard time explaining myself in the review.

Keep it up!

PR #275, opened 2023-02-19
Alexander Gabor - Week 2, Guess Who

Alexander-Gabor:
**Watch it live:**
https://whoisshegame.netlify.app/

Week 2: Guess Who

“Who is SHE?” -  Inspired by many powerful women creatives who cemented their footprints in their respective fields and beyond.


Cheroptera (commented):
I'm so impressed of what you managed to achieve in such a short time, considering that you also have a day job! Well done! 
You've added so many cool features and I really like the theme and styling of your game. 
Since it's a lot of code, it really helps with your descriptive comments to make it easy to follow, so thank you so much for adding that! ;) 
I love the timer and the count limit (even though I had a hard time beating this game). 
A few thoughts: 
- Would it be possible for the timer to start the first time the player hits the Find out button instead of directly when the page loads? 
- When you type in your name I understand that the name gets stored for later when you win or lose, but is it possible to either make the name input field disappear or present the name somehow on the page? 
Again, well done and can't wait to se what you create the coming weeks! I expect great things from you ;)

PR #278, opened 2023-02-19
Week 2 - Guess Who

SofiaGerdmar:
https://guess-who-green.netlify.app/


code/index.html
```diff
@@ -12,36 +12,54 @@
...
           <option value="hidden">hidden eyes</option>
-          <!-- Add the rest of eye colors as options  -->
+          <option value="brown">brown eyes</option>
+          <option value="green">green eyes</option>
+          <option value="blue">blue eyes</option>
+        </optgroup>
+        <optgroup label="hair">
+          <option value="hidden">hidden hair</option>
+          <option value="brown">brown hair</option>
+          <option value="blond">blond hair</option>
```
mvfrid, line 40: blonde, right? :)

code/index.html
```diff
@@ -52,11 +70,16 @@ <h1>Does the person have</h1>
...
           alt="Guess Who"
         />
         <h1 id="winOrLoseText"></h1>
-        <button type="button" id="playAgain" class="filled-button">PLAY AGAIN</button>
+        <button 
+          type="button" 
+          id="playAgain" 
+          class="filled-button">
+          PLAY AGAIN
+        </button>
```
mvfrid, line 82: I like this type of indentation, easy to read :)

code/script.js
```diff
@@ -205,7 +257,7 @@ let secret
 let currentQuestion
 let charactersInPlay
```
mvfrid, line 259: Seems to be working, so maybe no need to fix it, but most code I've seen has had the secret defined outside of the function, as a global variable.

code/script.js
```diff
@@ -222,104 +274,133 @@ const generateBoard = () => {
...
   secret = charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)]
 }
 
-// This function to start (and restart) the game
-const start = () => {
-  // Here we're setting charactersInPlay array to be all the characters to start with
-  charactersInPlay = CHARACTERS
-  // What else should happen when we start the game?
+// Lets the player confirm if they are sure before proceeding
+const guess = (secretPerson) => {
```
mvfrid, line 283: It's a bit confusing naming this argument "secretPerson", since it is actually the users guess. Now it sounds like it's the actual secret person, which makes it even more confusing when we are checking the guess in checkMyGuess function. Maybe rename it to userGuess, suspect, or similar?

code/script.js
```diff
@@ -1,202 +1,254 @@
...
     img: 'images/jabala.svg',
-    hair: 'hidden',
-    eyes: 'hidden',
-    accessories: ['glasses', 'hat'],
-    other: []
+    hairColor: 'hidden',
+    eyeColor: 'hidden',
+    glasses: true,
+    'a hat': true,
+    'smoking habit': false,
```
mvfrid, line 19: As we talked about in the demo, nice job on making the grammar work with the code!

mvfrid (approved):
Nice job on the responsiveness! Works really well. The only improvement I can see here is that the guess who logo is very small for mobile, even though the other parts are still roughly the same size as before. I think you could put it on the same row as the restart-button (use display flex maybe?), to make better use of the space. Or just size it up a liiiittle bit :) And maybe add some margin around the text on the win/lose page for mobile, as it looks a bit cramped.

Can we give user feedback that we have restarted the game when clicking restart (before the filtering)? Maybe by visually re-filling the board, with an alert, or another way? 

Overall, it looks nice and works like a dream. Great job Sofia!

PR #280, opened 2023-02-19
Week 2 Project "Guess who?"

cecavi:


code/script.js
```diff
@@ -222,104 +224,172 @@ const generateBoard = () => {
...
+      filterPeople(true)
+    } else {
+      filterPeople(false)
+    }
+    console.log("true or false")
   } else if (category === 'accessories' || category === 'other') {
-
+    if (secretPerson[category].includes(value)) {
+  }
+  filterPeople(true); {
```
SandraMadeleine, line 274: Between the filter people(true) and the filterpeople(false) there are too many curly brackets. Make sure that it looks just as you have done with filterPeople for the values hair and eyes (with the "else" also)

code/script.js
```diff
@@ -222,104 +224,172 @@ const generateBoard = () => {
...
-
-// when clicking guess, the player first have to confirm that they want to make a guess.
+//STEP 8 ----
+//when clicking guess, the player first have to confirm that they want to make a guess.
+ // If the player wants to guess, the checkMyGuess function is invoked.
 const guess = (personToConfirm) => {
-  // store the interaction from the player in a variable.
-  // remember the confirm() ?
-  // If the player wants to guess, invoke the checkMyGuess function.
+ confirmGuess= confirm(`Do you really want to make this guess?`)
```
SandraMadeleine, line 357: Works perfectly!

code/script.js
```diff
@@ -222,104 +224,172 @@ const generateBoard = () => {
...
+
+//STEP 9 ----
 // If you confirm, this function is invoked
-const checkMyGuess = (personToCheck) => {
   // 1. Check if the personToCheck is the same as the secret person's name
   // 2. Set a Message to show in the win or lose section accordingly
   // 3. Show the win or lose section
   // 4. Hide the game board
+
+const checkMyGuess = (personToCheck) => {
```
SandraMadeleine, line 372: This also works perfectly. For having the restart button showing you should not need to target it separately. I think that since it is a part of the whole section in the html and css for win or lose, it should show as a part of that. I remember struggling a bit with this one and ended up with a function just for this win or lose part, within the checkMyGuess function, which I had to call separately. I'm sure there's a much smoother way. :-)

SandraMadeleine (commented):
Good job Cecilia! This week was confusing and you did really well. :-) Since your game does filter for the hair color, it is probably just a small thing that needs to be adjusted in order for all the rest to filter as well. Maybe it was just one of those extra curly brackets in there? I think your idea of writing and/or re-writing everything yourself is smart, that is the best way to learn for sure!

PR #285, opened 2023-02-19
Project guess who

fannystenberg:


code/style.css
```diff
@@ -64,34 +71,38 @@ select {
...
 .card .guess {
   display: none;
   flex-direction: column;
   align-items: center;
   justify-content: center;
-  color: white;
+  color: #303599;
   text-align: center;
   height: inherit;
 }
```
AndreaHedstrom, line 102: Love that you have done it like the old (well sort of old) board game! With the colors and that the background on the cards is yellow when you hover it.

AndreaHedstrom (commented):
Great job with this project! I'm really impressed and a lot of things makes more sense for me when I read your code beacuse it's easy to follow and understand. Great with all the comments that describe what everything is, for example all the even listeners. 
Like I wrote in a comment I also really like your idea with the colors and background image that makes your game look like the older bord game version. And also the font and that the card is like "in front of the player" in the same way like in real life now when you got the buttons and text above instead. 

Didn't had much to add this week, the game works perfect as I can see and nothing is unclear in the code. GREAT job! 🐯

PR #294, opened 2023-02-19
Week 2 

edv-rd:


code/script.js
```diff
@@ -1,325 +1,452 @@
...
+  guessCounter.innerHTML = `Guesses: ${parseInt(numberOfGuesses)}`; // converting the int numberOfGuesses to a string. not sure this is needed actually?
+};
 
 // Randomly select a person from the characters array and set as the value of the variable called secret
 const setSecret = () => {
-  secret = charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)]
-}
+  secret =
+    charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)];
+  console.log("the person is set to " + secret.name);
```
smirrebinx, line 259: Maybe remove the console.log on line 259 since it reveals the secret person in the console :-)

code/script.js
```diff
@@ -1,325 +1,452 @@
...
-  const { category, value } = currentQuestion
+  const { category, value } = currentQuestion;
   // Show the correct alert message for different categories
-  if (category === 'accessories') {
+  if (category === "hair") {
     if (keep) {
-      alert(
-        `Yes, the person wears ${value}! Keep all people that wears ${value}`
-      )
+      addToLog(`The person has ${value} ${category}!`);
```
smirrebinx, line 344: A cool feature with the log, instead of alerts. 
It's not necessary but to clarify what happens in the game
you could add e.g. "remove all people with a hat" or "keep all people with a hat" in the log.

code/script.js
```diff
@@ -1,325 +1,452 @@
...
+      );
+    } else {
+      addToLog(`The person doesn't have ${value} ${category}!`);
+      charactersInPlay = charactersInPlay.filter(
+        (person) => person[category] !== value
+      );
+    }
+  } else if (category === "accessories") {
+    if (keep) {
+      addToLog(`The person wears ${value == "hat" ? "a" : ""} ${value}!`);
```
smirrebinx, line 368: When the player guesses on glasses it returns "The person does not wear false glasses". 
You could fix this by changing the option value in the CSS to "a hat" 
and then also change the key value in the array CHARACTERS to "a hat" instead of adding 
the a in the function which, in this case, returns false for the accessories that don't 
have an a in the array, such as glasses.

smirrebinx (changes requested):
Overall it looks really good! I like the guess-count and the guess-log. The detailed comments in the js-file made it clear what every function does.  The comment on 368 explains a suggested bug fix.

PR #299, opened 2023-02-19
Guess Who project week 2

SandraMadeleine:


code/script.js
```diff
@@ -229,97 +234,122 @@ const setSecret = () => {
...
 // Invokes the start function when website is loaded
-start()
+start();
 
 // All the event listeners
-restartButton.addEventListener('click', start)
+
+questions.addEventListener('change', selectQuestion); //when player chooses hair/eyes/accessories/other, the function selectQuestion is called
+findOutButton.addEventListener('click', checkQuestion); //when clicking on find out, the checkQuestion function is called
+restartButton.addEventListener('click', start); //restarts the game bt calling the function start
```
YlvaKarlsson, line 354: Nice job with the eventlisteners!
I also checked if the restart-button at your game works and it does! Both after some guesses or also if you guess on the wrong person. GREAT JOB!

I saw a minor thing that I think has to do with the CSS that you could fix if there's time (I'll try to post the printscreen here but if it doesn't work I'll send it on Slack instead :) ) 
image

After finishing the game - the game section is still there on the left side - I think you can chose the win-or-lose-section (or what you call it in your code) to cover the whole screen so that the game section is not visible. Hope that the pic I put her works so you see what I mean :)

YlvaKarlsson (approved):
Overall a very good job Sandra! 
You should be proud - your game works just fine, it filters characters away, it redraws the board and you can chose to play again.

I like the colors you choosed for your game but a tip for the future is to think how it looks an accesible way :)
Don't take it the wrong way - I mean nothing bad with this :) I like colors a lot but green and pink in a combination can be hard to read for a person with bad vision or eyesight so that can be good to have in mind for future projects :)

Really good job of you!

PS. I didn't get my game to even work so I'll look at your code while trying to fix mine...
See you soon again on Slack or in the Jitsi-room! :)

And don't forget to celebrate your progress :D

PR #300, opened 2023-02-19
Project Guess Who

EmmaEngvall:


code/script.js
```diff
@@ -222,104 +274,143 @@ const generateBoard = () => {
...
-  // 2. Set a Message to show in the win or lose section accordingly
-  // 3. Show the win or lose section
-  // 4. Hide the game board
+// Function to start (and restart) the game
+const start = () => {
+  charactersInPlay = CHARACTERS //reset character to original array
+  winOrLose.style.display = 'none' //dont show win or lose display 
+  board.style.display = 'flex' //show the board again
+  setSecret() //sets a new secret person
+  generateBoard() //draw board w all characters
```
code-and-cats, line 405: good explanations!

code-and-cats (approved):
Such a cute idea and nice implementation of the Guess who-project! The game works well and
I was genuinely excited when I got the banana!!

The code has a nice structure and is easy to follow thanks to good comments. 
I'm wondering working with arrays instead of booleans for accessories might make the code more comprehensive? The selectQuestion has a couple of conditionals based on attribute. Some could possibly be ommitted if arrays were used instead, and would lessen the ammount of properties of each object if something was to be changed later on.

I'd love to see a counter, but also know we are on a time constraint, so choices have to be made. I think it's a perfectly functioning game and it looks really cute with the Gru picture and the minions (I especially love Derek!). Well done!

PR #303, opened 2023-02-19
Oscar's Guess Who Project

OscarSindemark:
I realize the 'other' tab doesn't work correctly, have mercy on me!


RegSod (commented):
Neat coad. Easy to follow. Coldn't read the js-file though. When I click script js, I can only read CSS.

PR #307, opened 2023-02-22
Guess who app

dannebrob:
new PR, since the first only included the html?


EmmaEngvall:
Hi Daniel,

I think that you have done a great job with this project. You have really gone for the stretch goals with a question counter, timer and an overlay and executed them well. Good job!

One thing that I noticed is that no cards are removed when you chose an eye color. Perhaps that is something that you can double-check in your code. 

Apart from that, well done!

EmmaEngvall (commented):
Hi Daniel, 
Well done on the project!

Perhaps double check why no cards are removed when chosing a eye colour. 
But for now, be proud of the work that you have done!
/Emma

PR #312, opened 2023-09-08
project-guess-who-jennylarsen

jennylarsen:
Deployed version: https://guess-who-jennylarsen.netlify.app

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,4 +8,4 @@ With the help from my team, the material of the week, chat gpt and google I was 
 
 ## View it live
 
-
+https://guess-who-jennylarsen.netlify.app
--- a/README.md
+++ b/README.md
@@ -1,13 +1,11 @@
-# Project Name
+# Project Guess-who 
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+This project was about making a classical guess who game where a secret person is put by the computer and then the user starts guessing who it could be by asking the computer different statements to screen possible people to find the secret one. The focus of the project was to practice js and specifically objects, arrays and loops.
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+With the help from my team, the material of the week, chat gpt and google I was able to solve the project. My techniques was to break it down into the steps and try some code, see if it worked, changed it if it did not work and so on. If I had more time I would deep dive into the stretch goals. 
 
 ## View it live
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,13 +1,11 @@
...
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+This project was about making a classical guess who game where a secret person is put by the computer and then the user starts guessing who it could be by asking the computer different statements to screen possible people to find the secret one. The focus of the project was to practice js and specifically objects, arrays and loops.
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+With the help from my team, the material of the week, chat gpt and google I was able to solve the project. My techniques was to break it down into the steps and try some code, see if it worked, changed it if it did not work and so on. If I had more time I would deep dive into the stretch goals.
```
JuliaHolm, line 7: Good explanation!

code/script.js
```diff
@@ -2,6 +2,10 @@
 const board = document.getElementById('board')
 const questions = document.getElementById('questions')
 const restartButton = document.getElementById('restart')
+const findOutButton = document.getElementById('filter')
+const winOrLose = document.getElementById('winOrLose')
+const winOrLoseText = document.getElementById('winOrLoseText')
+const playAgainButton = document.getElementById('playAgain')
```
JuliaHolm, line 8: Good names, you can understand what everything is supposed to do. This applies for the whole code.

code/style.css
```diff
@@ -1,7 +1,7 @@
 /* Global css variables used for colors */
 :root {
-  --primary: #a259ff;
-  --secondary: #b0a6ff;
+  --primary: #F9E795;
+  --secondary: #F96167;
```
JuliaHolm, line 4: I love the new colors! Looks really nice.

JuliaHolm:
Everything works perfect! The code looks so clean so if I could I would write a comment on everything, but that's a bit to much so I decided to write it here instead 😄 If I was to point out something to add, it's one more option, a attribute that disable the first option in the dropdown and work as a placeholder to choose a question, as you can't ask if the person has brown hair when the game starts. But I see no problem not doing it.

- The board with characters should be generated when the website is loaded. ✅
- A randomly selected person should be set as the secret when the game starts. ✅
- You should be able to select questions to ask about the people and filter the board based on those questions. ✅
- You should also give the player feedback with alerts. ✅
- You should be able to guess the secret person and get an answer if it's correct or not. ✅

I guess you reached all the requirements, super duper job you done, I like it a lot! 👍

PR #315, opened 2023-09-08
project-guess-who-lauralyckholm

LauraLyckholm:
View it live:
https://guess-the-hp-character.netlify.app/

Problems:
I didn't really know how to solve the checkQuestion together with the filterQuestions function. I at first compared filterQuestion true or false inside nested if/else statements, which worked fine - but made the code longer than necessary. We where discussing the code in our team, which is where the thought of using the keep variable arose. By declaring keep within checkQuestion, that variable could instead be used to make the comparison. If the secret and value match that value is going to be assigned to "keep", so when the choice is "true" that alert will run from filterCharacters. I saved the function I made first inside the code as well. 

I wanted to add the new characters via an API, when I decided to try doing the stretch goals. I felt it would take more time to create my own new characters, than using an API and learning how to do that. I watched a video on FrontEnd Masters about it (Javascript First Steps with Anjana Vakil). But then, in the end, I still ended up putting the characters straight into the characters object to save time.

When changing the characters to others, and also their attributes, this led to many new problems! For once, this time I didn't have to use the (includes) because of arrays, but instead had to solve how to show an attribute that is a boolean. Lot's of Googling, but also a lot of help from ChatGPT finally made is clearer for me.


code/script.js
```diff
@@ -219,107 +230,142 @@ const generateBoard = () => {
...
-  /* 
-    for hair and eyes :
-      charactersInPlay = charactersInPlay.filter((person) => person[attribute] === value)
-      or
-      charactersInPlay = charactersInPlay.filter((person) => person[attribute] !== value)
+  if (category === 'house' || category === 'hair' || category === 'type' || category === 'species') {
+    filterFunction = person => person[category] === value;
+  } else if (category === 'teacher') {
+    filterFunction = person => person.isTeacher;
+  }
```
InnaKuljanac, line 293: This is great! What I did on serveral lines of code you did on just a few lines. It makes your code look so clean, good job!

code/script.js
```diff
@@ -219,107 +230,142 @@ const generateBoard = () => {
...
+    alert(
+      `Yes, the character ${category === 'teacher' ? 'is a teacher' : `has ${value} ${category}`}! Keep all matching characters`
+    );
+    charactersInPlay = positiveCharacters;
+  } else {
+    alert(
+      `No, the character ${category === 'teacher' ? `isn't a teacher` : `doesn't have ${value} ${category}`}! Remove all matching characters`
+    );
+    charactersInPlay = negativeCharacters;
+  }
```
InnaKuljanac, line 310: Just a tiny note, the alert doesn't always work with the options in your dropdown. For example if the player guesses on Hufflepuff house they get the alert "No, the character doesn't have hufflepuff house!". But this is just a minor detail!

InnaKokic (commented):
First of all, LOVE your theme!! The styling is really good and the game looks great!

As for the code, I think it looks great and I love how organised you are. From what I can see the game works exactly the way it should, and great job on the stretch goal with the counter!

code/script.js
LauraLyckholm: Thank you for this! I will look into it! I've had som issuses with this, but each time I do a push I have tried checking so that it works, and those times it does work. I'm a bit perplexed as too where the issue may lay :D

LauraLyckholm: Just a follow-up question, are you certain that it wrongly alerted No on Hufflepuff? I only have one person from Hufflepuff so maybe it just isn't visible that it disappears :D ?

InnaKuljanac: Oh no, I'm so sorry for being so unclear! The alert function works just fine!! I just meant that the sentence it puts out gets a little weird for some of the options. Like when you ask about a house it says ".. doesn't have Hufflepuff house" but it should maybe be something like ".. is not in hufflepuff". Hope you understand what I mean, like I said it's just a minor detail! :D

LauraLyckholm: Haha, oh ok, now I understand!! :D

PR #322, opened 2023-09-08
Project-guess-who-Janice

jforjanice:
# Project Guess Who

A guess who game developed by html, css and java script. Using filter function to compare if the selected character has the same attributes as the secret character in order to make the best guess.

## The problem

I thought I could have less struggle compare to the last project, but I found it very hard. I realised I still have some confusion about arrow function and how to passing functions as arguments. 
Also I spent a lot of time to understand the provided code. However I did realise I have improved a lot in terms of spotting my own issue.
By using a lot more console.log did help me to avoid a lot of mistakes, since I tried to make sure the variables work first before I move on and write more. 

## View it live
https://jocular-queijadas-c57f9a.netlify.app

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -311,7 +311,6 @@ const filterCharacters = (keep) => {
       charactersInPlay = charactersInPlay.filter((person) => person[category] !== value)
     }
   }
-
   else if (category === 'accessories') {
     if (keep) {
       alert(`Yes, the person wears ${value}! Keep all people that wears ${value}`)
@@ -321,6 +320,16 @@ const filterCharacters = (keep) => {
       charactersInPlay = charactersInPlay.filter((person) => !person[category].includes(value))
     }
   }
+  else if (category === 'other') {
+    if (keep) {
+      alert(`Yes, the person is a ${value}! Keep all people that is a ${value}`)
+      charactersInPlay = charactersInPlay.filter((person) => person[category].includes(value))
+    } else {
+      alert(`No, the person isn't a ${value}! Remove all people isn't a ${value}`)
+      charactersInPlay = charactersInPlay.filter((person) => !person[category].includes(value))
+    }
+  }
+  generateBoard(charactersInPlay)
   // I encounter some problems of the above block of code, I used chatGPT to check what was wrong. It provided an improvement of my code
   // Although I realized the problem was the value in the drop down list (hats) didn't match the attribube in the object (hat)
   // const filterCondition = keep
@@ -336,16 +345,6 @@ const filterCharacters = (keep) => {
 
   //   charactersInPlay = charactersInPlay.filter(filterCondition);
   // }
-
-
-  else if (category === 'other') {
-    if (keep) {
-      alert(`Yes, the person is a ${value}! Keep all people who smoke!`)
-    } else {
-      alert(`No, the person is not a ${value}! Remove all people who smoke!`)
-    }
-  }
-  generateBoard(charactersInPlay)
 
 }
 
--- a/code/script.js
+++ b/code/script.js
@@ -2,6 +2,8 @@
 const board = document.getElementById('board')
 const questions = document.getElementById('questions')
 const restartButton = document.getElementById('restart')
+const filterBtn = document.getElementById('filter')
+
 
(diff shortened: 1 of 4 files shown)


code/script.js
```diff
@@ -215,111 +220,197 @@ const generateBoard = () => {
...
   // Compare the currentQuestion details with the secret person details in a different manner based on category (hair/eyes or accessories/others).
   // See if we should keep or remove people based on that
   // Then invoke filterCharacters
-  if (category === 'hair' || category === 'eyes') {
+  // I created a group of console.log to help me check what information it contains and if I can use it in this scope
+  // console.log(secret.name)
+  // console.log(currentQuestion)
+  // console.log(charactersInPlay)
+  // console.log(secret[category])
+  // console.log(value)
```
Hang-Nguyen-Vu, line 272: Smart with the console.log, I suggest that you remove them before hand-in.

code/script.js
```diff
@@ -215,111 +220,197 @@ const generateBoard = () => {
...
     if (keep) {
-      alert(
-        `Yes, the person wears ${value}! Keep all people that wears ${value}`
-      )
+      alert(`Yes, the person has ${value}! Keep all people with ${value}`)
     } else {
-      alert(
-        `No, the person doesn't wear ${value}! Remove all people that wears ${value}`
-      )
+      alert(`No, the person doesn't have ${value}! Remove all people that doesn't have ${value}`)
```
Hang-Nguyen-Vu: The message should be "remove everyone that has....".

code/script.js
```diff
@@ -215,111 +220,197 @@ const generateBoard = () => {
...
     }
-  } else if (category === 'other') {
-    // Similar to the one above
-  } else {
+    charactersInPlay = charactersInPlay.filter((person) => person[category] === value === keep)
+  }
+  else if (category === 'eyes') {
     if (keep) {
-      // alert popup that says something like: "Yes, the person has yellow hair! Keep all people with yellow hair"
+      alert(`Yes, the person has ${value} color! Keep all people that has ${value} color`)
```
Hang-Nguyen-Vu: You could also print the category in the message, so you have e.g. green eyes etc

code/script.js
```diff
@@ -215,111 +220,197 @@ const generateBoard = () => {
...
+// It'll filter the characters array and redraw the game board.<
+// Use !== 
 const filterCharacters = (keep) => {
   const { category, value } = currentQuestion
   // Show the correct alert message for different categories
-  if (category === 'accessories') {
+  if (category === 'hair') {
     if (keep) {
-      alert(
-        `Yes, the person wears ${value}! Keep all people that wears ${value}`
```
Hang-Nguyen-Vu, line 272: Seems like you need to inject the category as well so that the message shows "black hair" instead of only "back".

code/script.js
```diff
@@ -215,111 +220,197 @@ const generateBoard = () => {
...
+// 4. Hide the game board
+
 const checkMyGuess = (personToCheck) => {
-  // 1. Check if the personToCheck is the same as the secret person's name
-  // 2. Set a Message to show in the win or lose section accordingly
-  // 3. Show the win or lose section
-  // 4. Hide the game board
+  if (personToCheck === secret.name) {
+    winOrLose.style.display = "flex";
+    winOrLoseText.innerHTML += `
```
Hang-Nguyen-Vu: I suggest writing only "=" here as you would like to replace the text with a new one depending on the result, not adding to it. I got once both win and lose messages when I guessed right.

code/style.css
```diff
@@ -1,7 +1,7 @@
 /* Global css variables used for colors */
 :root {
```
Hang-Nguyen-Vu, line 2: Great that you changed the background color, you could also change the text color of the value in the question to make it more visible

Hang-Nguyen-Vu (commented):
The codes are clear and structured and the game went quite well. Very descriptive comments which helped me understand the problem. If you could make some minor changes, it would be perfect! Overall good job!

code/script.js
jforjanice: Yes you are so right!!! Thanks !

PR #326, opened 2023-09-09
project-guess-who

BreannaSand:


code/script.js
```diff
@@ -2,6 +2,10 @@
 const board = document.getElementById('board')
 const questions = document.getElementById('questions')
 const restartButton = document.getElementById('restart')
+const findOutButton = document.getElementById('filter')
+const winOrLose = document.getElementById(`winOrLose`)
+const winOrLoseText = document.getElementById(`winOrLoseText`)
+const playAgainButton = document.getElementById(`playAgain`)
```
AmandaRam, line 8: As it is not a template string you can use single quotation marks instead :) Just a styling thing :)

code/script.js
```diff
@@ -225,27 +230,33 @@ const generateBoard = () => {
 // Randomly select a person from the characters array and set as the value of the variable called secret
 const setSecret = () => {
   secret = charactersInPlay[Math.floor(Math.random() * charactersInPlay.length)]
+  console.log(secret);
```
AmandaRam, line 233: Should not log the secret character as it is the users job to find out.

code/index.html
```diff
@@ -1,64 +1,66 @@
...
 
-      <button id="filter" class="filled-button">FIND OUT</button>
-    </aside>
+    <h1>Does the person have</h1>
 
-    <section class="board-wrapper">
-      <main class="game-board" id="board"></main>
-    </section>
+    <select title="questions" id="questions">
+      <option value="" disabled selected>--select an option:--</option>
```
AmandaRam, line 23: Nice with this set as default!

code/script.js
```diff
@@ -225,27 +230,33 @@ const generateBoard = () => {
...
 }
 
 // setting the currentQuestion object when you select something in the dropdown
 const selectQuestion = () => {
   const category = questions.options[questions.selectedIndex].parentNode.label
-
   // This variable stores what option group (category) the question belongs to.
   // We also need a variable that stores the actual value of the question we've selected.
-  // const value =
+  const value = questions.options[questions.selectedIndex].value
```
AmandaRam, line 253: Valid way of doing it, but there is a short hand for it:
```
const value = questions.value;
```
As the select always has a value attached to it (at least if the user has selected the value).

code/script.js
```diff
@@ -225,27 +230,33 @@ const generateBoard = () => {
...
 
   currentQuestion = {
     category: category,
-    // value: value
+    value: value
   }
+  console.log(currentQuestion);
```
AmandaRam, line 259: No need to log currentQuestion.

code/script.js
```diff
@@ -256,70 +267,117 @@ const checkQuestion = () => {
...
+    console.log("remove")
 
   } else if (category === 'accessories' || category === 'other') {
-
+    if (secret[category].includes(value)) {
+      filterCharacters(true)
+      console.log("keep")
+    } else
+      filterCharacters(false)
+    console.log("remove")
```
AmandaRam, line 283: No need to log keep and remove, but super nice use of array notation to get the desired property!

code/script.js
```diff
@@ -256,70 +267,117 @@ const checkQuestion = () => {
...
 }
 
-// when clicking guess, the player first have to confirm that they want to make a guess.
+// when clicking guess, the player first has to confirm that they want to make a guess.
 const guess = (personToConfirm) => {
   // store the interaction from the player in a variable.
   // remember the confirm() ?
+  const confirmation = confirm(`Are you sure you want to guess ${personToConfirm}?`)
   // If the player wants to guess, invoke the checkMyGuess function.
+  if (confirmation === true) {
```
AmandaRam, line 357: A short hand to write this is `if (confirmation)`.

code/script.js
```diff
@@ -256,70 +267,117 @@ const checkQuestion = () => {
...
-  // 2. Set a Message to show in the win or lose section accordingly
   // 3. Show the win or lose section
+  winOrLose.style.display = 'flex';
   // 4. Hide the game board
+  board.style.display = 'none';
+  // 2. Set a Message to show in the win or lose section accordingly
+  if (secret.name === personToCheck) {
+    winOrLoseText.innerHTML = `You guessed right! Congratulations, you are a winner!`
+  } else {
+    winOrLoseText.innerHTML = `You lost! The correct person was ${secret.name}.`
```
AmandaRam, line 372: You can use innerText instead of innerHTML as you are not setting HTML.

AmandaRam (approved):
There were only some styling things that cold be fixed, but not required!

Super nice work!! The code was easy to read and had a really good structure! I got inspired by your checkQuestion function as your code was shorter and smoother than mine!

PR #328, opened 2023-09-10
Wk 3 - Project Guess Who - Submitted

BeckieMorton:


code/script.js
```diff
@@ -222,104 +284,178 @@ const generateBoard = () => {
...
-// setting the currentQuestion object when you select something in the dropdown
+// Function - selectQuestion - setting currentQuestion object when you select something in the dropdown
 const selectQuestion = () => {
-  const category = questions.options[questions.selectedIndex].parentNode.label
 
-  // This variable stores what option group (category) the question belongs to.
-  // We also need a variable that stores the actual value of the question we've selected.
-  // const value =
+  const category = questions.options[questions.selectedIndex].parentNode.label //stores selected category after find out button
+  const value = questions.options[questions.selectedIndex].value //stores selected value after find out button
```
sansan-sakura, line 299: This is a good choice!! I was really stupid, and I chose to go with text instead of value... It went with a lot of obstacles.

code/script.js
```diff
@@ -222,104 +284,178 @@ const generateBoard = () => {
...
 
+  //declare an object
   currentQuestion = {
     category: category,
-    // value: value
+    value: value
   }
+  return (currentQuestion) //do I need this??
```
sansan-sakura, line 306: You don't need to return here, because you assigned to currentQuestion which has been declared at the beginning of your code.  As far as I know, when you need to return to something is when you need to get something through a function.  
1, A function returns undefined, if you don't return.
2, Arrow function might return without writing "return", if it is a just single line of code.
`const calcAge=birthYear=> 2023-birthYear` `calcAge(1999)` -----> 24
The function above doesn't need to return, but it returns automatically. 
### This function needs to be returned, as it produces something and doesn't assign it to a variable.  ####
So if you want to use function declaration, you need to write return.

I hope it makes sense to you...

code/script.js
```diff
@@ -222,104 +284,178 @@ const generateBoard = () => {
...
-      charactersInPlay = charactersInPlay.filter((person) => person[attribute] !== value)
-
-    for accessories and other
-      charactersInPlay = charactersInPlay.filter((person) => person[category].includes(value))
-      or
-      charactersInPlay = charactersInPlay.filter((person) => !person[category].includes(value))
-  */
-
-  // Invoke a function to redraw the board with the remaining people.
 }
```
sansan-sakura, line 395: I found your code is quite organized and easy to read! And I noticed one small thing that you can even make it better.  You can call just generateBored function once at the end of filterFunction, so don't need to place it many times:))

code/script.js
```diff
@@ -222,104 +284,178 @@ const generateBoard = () => {
...
-  // store the interaction from the player in a variable.
-  // remember the confirm() ?
-  // If the player wants to guess, invoke the checkMyGuess function.
+
+  let confirmGuess = confirm(`Are you sure you want to guess on ${personToConfirm}?`);
+
+  if (confirmGuess) {
+    checkMyGuess(personToConfirm);
+  } else if (!confirmGuess) {
+    return;
```
sansan-sakura, line 406: Here you don't need to return it, if you want. You can just write one statement and no else if.

sansan-sakura (commented):
Overall, I found your code very organised and nicely commended!  Well done! I also like the colour you chose. 

One thing I think you need to fix is your timer!!! It seems to work the first time I play, and when I click the second time it goes too fast I think.

PR #342, opened 2023-09-10
Guess Who: Middle-Earth Quest / Olga

olgalepisto:
This is my Lord of the Rings themed guess who project called Guess Who: Middle-Earth Quest!
Your job is to guess the right character to save Middle-Earth.

Link to the deployed site: https://middle-earth-quest.netlify.app/

Note: Unfortunately the html audio player I put to the corner does not work with the deployed site :(
So it is highly suggested that you listen to some LOTR tunes while playing this game.


colognia (commented):
Good pull request description and the code does not have bugs.

- player not working, mp3 path seems to be wrong
- elements in aside not all visible in viewport, because of big gap
- For the commits the description should be more informative and containing a body about details -> https://cbea.ms/git-commit/

- all tasks from 1-5 are fullfilled
- bonustask: working restart button, properties of items changed
- added new value to dropdown for user choice
- nice extensive costum styling to make it unique

PR #343, opened 2023-09-10
Guess who - game

FridaKristian:


FridafridaL (commented):
Frida! Great work with your 'Guess who'-project. The code looks nice and clear. I like that you manage to make the code short, fort example in the selectQuestion-function. But still understandable. The game works as it should in the browser and the fallout can be both winning or losing :)
You've also made a guess-counter which was in the stretch goals. Thumbs up!

PR #348, opened 2023-09-11
Gues who

hamdiolad:


code/script.js
```diff
@@ -2,8 +2,14 @@
 const board = document.getElementById('board')
 const questions = document.getElementById('questions')
 const restartButton = document.getElementById('restart')
+const winOrLose = document.getElementById('winOrLose')
+const winOrLoseText = document.getElementById('winOrLoseText')
+const playAgainButton = document.getElementById('playAgain')
```
BreannaSand: Where is this button's event listener?

code/script.js
```diff
@@ -2,8 +2,14 @@
...
+const findOutButton = document.getElementById('filter')
```
BreannaSand: This button also needs an event listener.

code/script.js
```diff
@@ -2,8 +2,14 @@
 const board = document.getElementById('board')
 const questions = document.getElementById('questions')
```
BreannaSand: The dropdown menu needs an event listener.

code/script.js
```diff
@@ -220,106 +227,177 @@ const generateBoard = () => {
...
+      person[category].includes(value)
+    );
+  } else {
+    charactersInPlay = charactersInPlay.filter(
+      (person) => !person[category].includes(value)
+    );
+  }
 }
+// Invoke a function to redraw the board with the remaining people.
+generateBoard(keep);}
```
BreannaSand: I am guessing that your board isn't generating at the startup of the game when this function is called because "keep" doesn't have a value. I could be wrong as I can't get to that part in the code when I run it but when I try following the logic I don't see where the variable "keep" is ever actually assigned a value.

code/script.js
```diff
@@ -220,106 +227,177 @@ const generateBoard = () => {
...
+};
+
+
 
 // Invokes the start function when website is loaded
 start()
 
 // All the event listeners
 restartButton.addEventListener('click', start)
+find
```
BreannaSand: I'm not sure that this needs to be here.

code/script.js
```diff
@@ -220,106 +227,177 @@ const generateBoard = () => {
...
-      )
+      matchingCharacters.push(character);
     }
-  } else if (category === 'other') {
-    // Similar to the one above
+    filterCharacters(keep, matchingCharacters);
+
+  };
+
+  const filterCharacters = (keep, matchingCharacters) => {
```
BreannaSand: There are two filterCharacter functions, when I plug your code into VScode it shows several functions that don't have closing curly brackets. Both filterCharacter functions and checkQuestion function.

code/script.js
```diff
@@ -220,106 +227,177 @@ const generateBoard = () => {
...
   // remember the confirm() ?
   // If the player wants to guess, invoke the checkMyGuess function.
-}
+  if (confirmGuess) {
+    checkMyGuess(personToConfirm);
+  }
+};
 
 // If you confirm, this function is invoked
 const checkMyGuess = (personToCheck) => {
```
BreannaSand, line 382: checkMyGuess function has an error in the syntax of the if statement.

BreannaSand (changes requested):
Hey Hamdi! Great job getting back up to speed after losing all your work! It seems like your code is mostly there, it's just to keep playing around with the small bugs until you get it working. I identified some things that you can start with and from there it is hard to predict. I wish I could help you more so please write to me if you have any questions during your troubleshooting, I'd be happy to help if I can! Right now it's a little difficult to give any more without spending a lot of time because the webpage is only showing the interface of the starter code without any functionality and no board. Hopefully if you figure out the generateBoard thing I commented on it will make the rest easier to fix. Good luck! :)

